PR #13, opened 2020-05-28
Fix double hash and implement isValidSignature

Agusx1211:
- Fixes double hashing during `Wallet.signMessage`
- Implements `isValidSignature` with compatibility for EIP712, EthSign, Wallet sign, Wallet sign (Arcadeum), Arcadeum static sign (counterfactual against create2)

Implements https://github.com/horizon-games/issue-tracker/issues/2700
Replaces https://github.com/arcadeum/arcadeum.js/pull/11

### DIFF
--- a/packages/provider/src/utils.ts
+++ b/packages/provider/src/utils.ts
@@ -128,19 +128,20 @@ export async function isValidSignature(
 
   const wallets = await Promise.all([
     isValidWalletSignature(address, digest, sig, provider),
-    isValidArcadeumWalletSignature(address, digest, sig, provider, chainId)
+    isValidArcadeumDeployedWalletSignature(address, digest, sig, provider, chainId)
   ])
 
   // If validity of wallet signature can't be determined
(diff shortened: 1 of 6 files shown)


packages/provider/src/utils.ts
```diff
@@ -112,10 +113,123 @@ function recoverSigner(digest: Arrayish, sig: ArcadeumDecodedSigner) {
...
+
+export async function isValidArcadeumWalletSignature(
+  address: string,
+  digest: Uint8Array,
+  sig: string,
+  provider?: Provider
+) {
+  if (!provider) throw new Error('Arcadeum wallet signatures require RPC Provider')
+  try {
+    const chainId = (await provider.getNetwork()).chainId
```
PhABC, line 197: What if provider is on mainnet but we also want to verify on, say, POA or another network?

Agusx1211: Yep! Good catch

PR #14, opened 2020-05-29
Implement updateConfig method

Agusx1211:
- Add `updateConfig` method to wallet
-> Migrate wallet from `fixed` to `upgradable` if required
-> Use bundle atomicity for secure migration
-> Detect and reject non-usable configurations
- Add fixed `address` as an optional parameter in configuration
- Add strict mode on context, used to determined if non-usable configurations should be rejected
-> Detect and reject non-usable configurations during wallet creation
- Upgrade contracts

note: non-usable configurations are valid, but are not usable because the sum of all signers doesn't reaches the threshold.

### DIFF
--- a/packages/provider/src/utils.ts
+++ b/packages/provider/src/utils.ts
@@ -34,6 +34,8 @@ export function sortConfig(config: ArcadeumWalletConfig): ArcadeumWalletConfig {
   return config
 }
 
+// isUsableConfig checks if a the sum of the owners in the configuration meets the necessary threshold to sign a transaction
+// a wallet that has a non-usable configuration is not able to perform any transactions, and can be considered as destroyed
 export function isUsableConfig(config: ArcadeumWalletConfig): boolean {
   const sum = config.signers.reduce((p, c) => ethers.utils.bigNumberify(c.weight).add(p), ethers.constants.Zero)
   return sum.gte(ethers.utils.bigNumberify(config.threshold))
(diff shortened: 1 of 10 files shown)


packages/provider/src/utils.ts
```diff
@@ -34,6 +34,11 @@ export function sortConfig(config: ArcadeumWalletConfig): ArcadeumWalletConfig {
   return config
 }
 
+export function isUsableConfig(config: ArcadeumWalletConfig): boolean {
```
pkieltyka, line 37: what does the term "usable" mean? perhaps you could leave a comment above the function line to describe its use and scenario?

Agusx1211: Non-usable means that's valid but the sum of the signers is below the required threshold, I wanted to avoid the 'invalid' word, because the configuration is "valid" in some way.

I can add a comment, or we can change the nomenclature and consider configurations that don't reach the threshold to be "invalid"

pkieltyka: I think the "usable" sounds good -- just leaving a comment above as you mentioned is perfect. Keep it simple, in Go, a lot of people just leave a single comment in form or `// <function> ...` for example, `// isUsableConfig checks if a wallet configuration meets the necessary threshold to perform successful signing / transactions. A usable config can perform signing, and an unusable config means the user requires to add another signer to meet minimum security requirements.`

PR #24, opened 2020-07-02
Estimate gasLimits locally

Agusx1211:
- Estimate gas limit using local-relayer
- When gas can't be estimated (self-calls, delegate-calls, etc) use default gasLimit
- Auto-estimate gas limit if all transactions have gasLimit 0 and revertOnError false

### DIFF
--- a/packages/provider/src/relayer/local-relayer.ts
+++ b/packages/provider/src/relayer/local-relayer.ts
@@ -9,6 +9,8 @@ import { addressOf } from '../utils'
 
 import { IRelayer } from '.'
 
+const DEFAULT_FEE = ethers.utils.bigNumberify(800000)
+
 export class LocalRelayer extends BaseRelayer implements IRelayer {
   private readonly signer: Signer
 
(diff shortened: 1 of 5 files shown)


packages/provider/src/relayer/local-relayer.ts
```diff
@@ -9,6 +9,8 @@ import { addressOf } from '../utils'
 
 import { IRelayer } from '.'
 
+const DEFAULT_FEE = ethers.utils.bigNumberify(800000)
```
pkieltyka, line 12: can we rename this to, "DEFAULT_GAS_LIMIT"? as this isn't really the fee, its just the limit? fee would be price*limit

packages/provider/src/relayer/local-relayer.ts
```diff
@@ -23,6 +25,40 @@ export class LocalRelayer extends BaseRelayer implements IRelayer {
...
+  ): Promise<ArcadeumTransaction[]> {
+    const walletAddr = addressOf(config, context)
+
+    const gasCosts = await Promise.all(transactions.map(async (t) => {
+      // Fee can't be estimated locally for delegateCalls
+      if (t.delegateCall) {
+        return DEFAULT_FEE
+      }
+
+      // Fee can't be estimated for self-called if wallet hasn't been deployed
```
pkieltyka, line 41: how come we can't estimate a self-called wallet that hasn't been deployed?

packages/provider/src/relayer/local-relayer.ts
Agusx1211: Yep, good catch

packages/provider/src/relayer/local-relayer.ts
Agusx1211: If the wallet is a self-call we need to have the wallet code deployed, otherwise the estimation call will be performed to a wallet without code and no "real" gas estimation will take place.

PR #27, opened 2020-07-16
RPC gas estimation

Agusx1211:
- [x] Implement `estimateGasLimits` on rpc-relayer
- [x] Implement `gasRefundOptions`
- [ ] Add option to append `gasRefundOptions` on provider ?


packages/provider/tests/wallet.e2e.spec.ts
```diff
@@ -52,70 +52,202 @@ if (process.env.ONLY_E2E) {
...
+            to: callReceiver.address,
+            value: 0,
+            data: callReceiver.interface.functions.testCall.encode([7435, '0x445566'])
+          }]
 
-        const tx2 = await wallet.sendTransaction(transaction2)
-        await tx2.wait()
+          const arctxs = await toArcadeumTransactions(wallet, transactions, true)
+          const estimated = await relayer.estimateGasLimits(wallet.config, ARCADEUM_CONTEXT, ...arctxs)
+          const tx = await wallet.sendTransaction(estimated)
```
pkieltyka, line 183: the estimation stuff looks great, but where do we include the token fee from the client side..? that is, when a user is presented to sign a txn, we should also include the txn for the fee token, based on the estimate, so the user will sign the batch, etc..? is this included or does the wallet code need to include the fee txn into the batch upon confirmation?

Agusx1211: I think so, the wallet should take the batch of transactions and pass them through `estimateGasLimits`, then those transactions can be sent to `gasRefundOptions` to get a list of all possible refund options

PR #31, opened 2020-08-12
Add web3 providers for sidechains

Agusx1211:
Adds support for injecting sidechains web3 providers.

### DIFF
--- a/packages/provider/src/providers/external-window-provider.ts
+++ b/packages/provider/src/providers/external-window-provider.ts
@@ -20,7 +20,7 @@ export class ExternalWindowProvider implements AsyncSendable {
   private networkPayload?: NetworkConfig
   private events: EventEmitter<EventType, any> = new EventEmitter()
 
-  constructor(walletAppURL: string, chainId?: number) {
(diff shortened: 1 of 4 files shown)


packages/provider/src/providers/external-window-provider.ts
```diff
@@ -20,7 +20,7 @@ export class ExternalWindowProvider implements AsyncSendable {
   private networkPayload?: NetworkConfig
   private events: EventEmitter<EventType, any> = new EventEmitter()
 
-  constructor(walletAppURL: string) {
+  constructor(walletAppURL: string, chainId?: number) {
```
pkieltyka, line 23: how come we need to pass chainId to the ExternalWindowProvider constructor?

pkieltyka: looks like you're not using the argument in fact, so probably an early idea and we can remove it now -- which makes more sense since the ExternalWindowProvider would be chainId agnostic, it would be the requests that would be directed at a specific chainId

PR #32, opened 2020-08-13
Expose batch transactions on window provider

Agusx1211:
Add `getSequenceSigner` to injected provider.
Sequence signer enables batch expirable transactions using `sendTransactionBatch`

### DIFF
--- a/packages/provider/src/context.ts
+++ b/packages/provider/src/context.ts
@@ -4,7 +4,8 @@ export const WalletContext: ArcadeumContext = {
   factory: '0x52f0F4258c69415567b21dfF085C3fd5505D5155',
   mainModule: '0x57bA6Eb1ed6821Db7bC26C6714f093E2BCbea40F',
   mainModuleUpgradable: '0xF52136A04057d889CeBf9bCc2F3142965AeABc54',
-  guestModule: '0xe076ad01F1eb18A8eF20bB64003DA4810a429a32'
+  guestModule: '0xe076ad01F1eb18A8eF20bB64003DA4810a429a32',
+  requireUtils: '0xd46566c23ae549fc14abfcfcbc130e3ea7e63f82'
 }
 
 export const ethereumNetworks: {[key: string]: NetworkConfig} = {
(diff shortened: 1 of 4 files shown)


packages/provider/src/context.ts
```diff
@@ -4,7 +4,8 @@ export const WalletContext: ArcadeumContext = {
   factory: '0x52f0F4258c69415567b21dfF085C3fd5505D5155',
   mainModule: '0x57bA6Eb1ed6821Db7bC26C6714f093E2BCbea40F',
   mainModuleUpgradable: '0xF52136A04057d889CeBf9bCc2F3142965AeABc54',
-  guestModule: '0xe076ad01F1eb18A8eF20bB64003DA4810a429a32'
+  guestModule: '0xe076ad01F1eb18A8eF20bB64003DA4810a429a32',
+  requireUtils: '0xd46566c23ae549fc14abfcfcbc130e3ea7e63f82'
```
pkieltyka, line 8: what is this contract about btw?

packages/provider/src/providers/sequence-web3-provider.ts
```diff
@@ -0,0 +1,61 @@
+import { Web3Provider, AsyncSendable, TransactionRequest, TransactionResponse, JsonRpcSigner } from "ethers/providers"
+import { Networkish, BigNumberish, Interface } from "ethers/utils"
+import { toArcadeumTransactions, makeExpirable, makeAfterNonce, arcadeumTxAbiEncode } from "../utils"
+import { NonceDependency, ArcadeumContext } from "../types"
+import { abi as mainModuleAbi } from '../abi/mainModule'
+import { ethers, Signer } from "ethers"
+
+export class SequenceWeb3Provider extends Web3Provider {
```
pkieltyka, line 8: let's call this still ArcadeumWeb3Provider, to be consistent with naming until we switch everything one-shot -- makes it less confusing I think

packages/provider/src/context.ts
Agusx1211: It's a very small contract that implements the functionality for "expirable" transactions and "linked" transactions. It has methods with only a "require" in them, so the wallet can include txs to those methods with `revertOnError: true` and with that enforce those conditions on the transaction batch.

PR #37, opened 2020-08-27
Implement MultiWallet multisig wallet

Agusx1211:
Implement `MultiWallet` a wallet class that can be used to track configuration changes across multiple chains, it's interface-compatible with `Wallet`.

### DIFF
--- a/packages/provider/src/multi-wallet.ts
+++ b/packages/provider/src/multi-wallet.ts
@@ -5,7 +5,8 @@ import { NetworkConfig, ArcadeumWalletConfig, ArcadeumContext, Transactionish } 
 import { Arrayish, BigNumberish } from 'ethers/utils'
 import { abi as mainModuleUpgradableAbi } from './abi/mainModuleUpgradable'
 import { abi as requireUtilsAbi } from './abi/requireUtils'
-import { isConfig, isAsyncSendable } from './utils'
+import { isConfig } from './utils'
+import { NotEnoughSigners } from './errors'
 
 // TODO: Add more details to network
 // authChain, mainNetwork, etc
@@ -125,7 +126,7 @@ export class MultiWallet extends AbstractSigner {
     }, ethers.constants.Zero)
 
     if (weight.lt(lastConfig.threshold)) {
-      throw Error('Not enough signing power for sending a transaction')
+      throw new NotEnoughSigners(`Wallet combined weight ${weight.toString()} below required ${lastConfig.threshold.toString()}}`)
     }
 
     // If the wallet is updated, procede to transaction send
--- a/packages/provider/src/multi-wallet.ts
+++ b/packages/provider/src/multi-wallet.ts
@@ -106,14 +106,12 @@ export class MultiWallet extends AbstractSigner {
     const wallet = network ? this.networkWallet(network) : this.mainWallet()
 
     // TODO: Skip this step if wallet is authWallet
-    const [thisConfig, lastConfig] = await Promise.all([
-      this.currentConfig(wallet), this.currentConfig()
-    ])
+    const thisConfig = await this.currentConfig(wallet)
 
     // See if wallet has enough signer power
     const weight = await wallet.useConfig(thisConfig).signWeight()
     if (weight.lt(thisConfig.threshold)) {
-      throw new NotEnoughSigners(`Sign message - wallet combined weight ${weight.toString()} below required ${lastConfig.threshold.toString()}`)
+      throw new NotEnoughSigners(`Sign message - wallet combined weight ${weight.toString()} below required ${thisConfig.threshold.toString()}`)
     }
 
     return wallet.useConfig(thisConfig).signMessage(message)
@@ -130,7 +128,7 @@ export class MultiWallet extends AbstractSigner {
     // See if wallet has enough signer power
     const weight = await wallet.useConfig(thisConfig).signWeight()
     if (weight.lt(thisConfig.threshold)) {
-      throw new NotEnoughSigners(`Send transaction - wallet combined weight ${weight.toString()} below required ${lastConfig.threshold.toString()}`)
+      throw new NotEnoughSigners(`Send transaction - wallet combined weight ${weight.toString()} below required ${thisConfig.threshold.toString()}`)
     }
 
     // If the wallet is updated, procede to transaction send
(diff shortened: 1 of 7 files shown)


packages/provider/src/multi-wallet.ts
```diff
@@ -0,0 +1,271 @@
...
+import { NetworkConfig, ArcadeumWalletConfig, ArcadeumContext, Transactionish } from './types'
+import { Arrayish, BigNumberish } from 'ethers/utils'
+import { abi as mainModuleUpgradableAbi } from './abi/mainModuleUpgradable'
+import { abi as requireUtilsAbi } from './abi/requireUtils'
+import { isConfig } from './utils'
+import { NotEnoughSigners } from './errors'
+
+// TODO: Add more details to network
+// authChain, mainNetwork, etc
+export type NetworkWallet = {
```
pkieltyka, line 13: btw, what is a "network wallet" ?

packages/provider/src/multi-wallet.ts
```diff
@@ -0,0 +1,271 @@
...
+
+  get address(): string {
+    return this._wallets[0].wallet.address
+  }
+
+  getAddress(): Promise<string> {
+    return this._wallets[0].wallet.getAddress()
+  }
+
+  async getFullConfig(): Promise<FullConfig> {
```
pkieltyka, line 76: what is a "full config" ? versus just a normal wallet config?

packages/provider/src/multi-wallet.ts
```diff
@@ -0,0 +1,271 @@
...
+      signers: signers.map((s: any) => ({
+        address: s.signer,
+        weight: s.weight
+      }))
+    }
+
+    return config
+  }
+
+  private networkWallet(network: NetworkConfig | BigNumberish): Wallet {
```
pkieltyka, line 251: what is a "network wallet" ?

packages/provider/src/multi-wallet.ts
Agusx1211: It's a wallet for a given network, each network has it's own "internal" wallet

packages/provider/src/multi-wallet.ts
Agusx1211: Because there are multiple networks, the wallet may have different configurations on each one of those networks. This method returns all the configuration of all the current networks.

packages/provider/src/multi-wallet.ts
Agusx1211: This is a helper method used to retrieve the wallet of a given network, (an internal wallet it's created for each one of the networks).

packages/provider/src/multi-wallet.ts
```diff
@@ -0,0 +1,271 @@
...
+  }
+
+  getAddress(): Promise<string> {
+    return this._wallets[0].wallet.getAddress()
+  }
+
+  async getFullConfig(): Promise<FullConfig> {
+    const allConfigs = await Promise.all(this._wallets.map(async (w) => ({ wallet: w, config: await this.currentConfig(w.wallet) })))
+    const thresholds = allConfigs.map((c) => ({ chaind: c.wallet.network.chainId, weight: c.config.threshold }))
+    const allSigerns = allConfigs.reduce((p, config) => {
```
PhABC, line 79: Typo, should be `allSigners`*

packages/provider/src/multi-wallet.ts
```diff
@@ -0,0 +1,271 @@
...
+        const item = p.find((c) => c.address === signer.address)
+        const netEntry = {
+          weight: signer.weight,
+          chaind: config.wallet.network.chainId
+        }
+
+        if (!item) {
+          p.push({
+            address: signer.address,
+            networks: [netEntry]
```
PhABC, line 90: why is the `weight` in the `networks` field? Seems a bit un-intuitive. Why not have three fields `address, network, weight`?

packages/provider/src/multi-wallet.ts
```diff
@@ -0,0 +1,271 @@
...
+    // TODO: Skip this step if wallet is authWallet
+    const [thisConfig, lastConfig] = await Promise.all([
+      this.currentConfig(wallet), this.currentConfig()
+    ])
+
+    // See if wallet has enough signer power
+    const weight = await wallet.useConfig(thisConfig).signWeight()
+    if (weight.lt(thisConfig.threshold)) {
+      throw new NotEnoughSigners(`Send transaction - wallet combined weight ${weight.toString()} below required ${thisConfig.threshold.toString()}`)
+    }
```
PhABC, line 132: Not sure I understand this `(weight.lt(thisConfig.threshold))` section, would you mind explaining?

packages/provider/src/multi-wallet.ts
Agusx1211: The address may or may not exist on different networks, and on each network the address could have a different weight. Having the 3 fields works to, but at that point it's practically an array of configurations. My idea here was to try to do a "flatten" version of the configuration since that's needed to show the devices list on the wallet webapp

packages/provider/src/multi-wallet.ts
Agusx1211: `weight` is the combined `weight` of all signers currently configured on the wallet, the `threshold` is the `threshold` of the current configuration. If the wallet doesn't have enough signers to sign a message/transaction then this method throws an error.

PhABC: Under what condition would a wallet not have enough signers? Would have to be an invalid config, no?

Agusx1211: This throws when the wallet instance has not enough accessible signers (private keys + remote signers), imagine a 2/3 wallet with signers A, B, and C. If you create an instance of `arcadeum.MultiWallet` with only A private key then you need either B or C to sign transactions.

One could argues that those signers exists somewhere else and the relayer is going to aggregate the signatures later, for that maybe we can add a flag to the method `revertOnCantSign` or similar

PhABC: Yeah I guess I always assumed we would go with this assumption:

> One could argues that those signers exists somewhere else and the relayer is going to aggregate the signatures later,

I.e. that local wallets won't have many keys on their own, unless in some exceptions.

Agusx1211: We can always include non-local signers in the list of accessible signers of `arcadeum.Wallet` and because the `signTransaction` and `signMessage` are both async we can request the signatures of remote signers in a transparent way. This would only work when the client knows who the remote signers is, and that's the case for the Arcadeum wallet and the guard signer.

PR #52, opened 2020-10-16
Preset sidechain provider cache with chain id

Agusx1211:


acrylix:
resolves https://github.com/horizon-games/issue-tracker/issues/3474

PR #61, opened 2020-12-17
multicall: review

pkieltyka:


packages/multicall/src/multicall.ts
```diff
@@ -1,16 +1,21 @@
...
-export type MulticallConf = {
+export type MulticallOptions = {
+  // number of calls to enqueue before calling.
   batchSize: number,
+
+  // number of calls to batch within a time window (in milliseconds). If 0, will disable timeWindow.
   timeWindow: number,
-  contract: string
+
+  // only batch calls for a specific contract.
```
Agusx1211, line 17: This contract is the address of the SequenceUtils contracts, maybe we could rename the variable

packages/multicall/src/multicall.ts
```diff
@@ -21,38 +26,39 @@ type QueueEntry = {
...
+  timeWindow: 50
+}
+
 export class Multicall {
-  public static DEFAULT_CONF = {
-    batchSize: 50,
-    timeWindow: 50,
-    contract: ""
-  }
+  public static DefaultOptions = () => DefaultMulticallOptions
```
Agusx1211, line 35: Why using an arrow function?

Agusx1211 (approved):

packages/multicall/src/multicall.ts
pkieltyka: ok, good to kno! I'll update

packages/multicall/src/multicall.ts
pkieltyka: just for style / short-hand

PR #69, opened 2020-12-23
Next sign and send

Agusx1211:
* Implement eth_signTransaction
* Implement eth_sendRawTransaction
* Create `SignedTransactions` type
* Add `getProvider` and `getRelayer` to Signer interface

### DIFF
--- a/packages/provider/src/wallet-request-handler.ts
+++ b/packages/provider/src/wallet-request-handler.ts
@@ -7,14 +7,14 @@ import {
   MessageToSign
 } from './types'
 
-import { ethers, Signer as AbstractSigner } from 'ethers'
-import { JsonRpcProvider, ExternalProvider, TransactionResponse } from '@ethersproject/providers'
+import { BigNumber, ethers } from 'ethers'
+import { JsonRpcProvider, ExternalProvider } from '@ethersproject/providers'
 import { TypedDataUtils } from 'ethers-eip712'
 
-import { Networks, NetworkConfig, JsonRpcHandler, JsonRpcRequest, JsonRpcResponseCallback, JsonRpcResponse, isJsonRpcProvider } from '@0xsequence/network'
+import { Networks, NetworkConfig, JsonRpcHandler, JsonRpcRequest, JsonRpcResponseCallback, JsonRpcResponse } from '@0xsequence/network'
 
 import { Signer } from '@0xsequence/wallet'
-import { SequenceTransaction, appendNonce, readSequenceNonce, toSequenceTransactions, isSequenceTransaction, flattenAuxTransactions } from '@0xsequence/transactions'
+import { isSignedTransactions } from '@0xsequence/transactions'
 
 export class WalletRequestHandler implements ExternalProvider, JsonRpcHandler, ProviderMessageRequestHandler {
   private wallet: Signer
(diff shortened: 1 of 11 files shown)


packages/wallet/src/signer.ts
```diff
@@ -1,16 +1,21 @@
...
+import { Relayer } from '@0xsequence/relayer'
+import { SequenceTransaction, SignedTransactions, Transactionish } from '@0xsequence/transactions'
+import { Provider, TransactionResponse } from '@ethersproject/providers'
 import { BigNumberish, Signer as AbstractSigner } from 'ethers'
 import { BytesLike, Deferrable } from 'ethers/lib/utils'
 import { WalletConfig } from '.'
 
 export abstract class Signer extends AbstractSigner {
-  readonly provider?: JsonRpcProvider
+  abstract getProvider(chainId?: number): Promise<Provider | undefined>
```
pkieltyka, line 10: can we return Promise<JsonRpcProvider | undefined> instead? I think will be easier to consume and why not use JsonRpcProvider for all?

packages/provider/src/wallet-request-handler.ts
```diff
@@ -177,70 +177,35 @@ export class WalletRequestHandler implements ExternalProvider, JsonRpcHandler, P
...
             // ...
           }
 
-          const sig = signer.signTransaction(transaction)
-          response.result = {
-            raw: sig,
-            tx: transaction
+          if (sender === await signer.getAddress()) {
+            // const signed = await signer.signTransactions(transaction)
+            response.result = await signer.signTransactions(transaction)
```
pkieltyka, line 190: Is the return type of signTransactions in the format that eth_signTransaction json-rpc response type?

packages/wallet/tests/wallet.spec.ts
```diff
@@ -829,8 +833,8 @@ describe('Wallet integration', function () {
         const signed_3 = await w3_3.eth.signTransaction(transaction)
 
         const full_signed = {
-          raw: aggregate(signed_1.raw, signed_2.raw, signed_3.raw), // TODO: 'aggregate' name is too vague
-          tx: signed_1.tx
+          ...signed_1,
+          signature: aggregate(signed_1.signature, signed_2.signature, signed_3.signature) // TODO: 'aggregate' name is too vague
```
pkieltyka, line 837: how about `joinSignatures()` ? .. and btw, what kind of "signatures" are these? maybe we can make it more specific then, ie. `joinTransactionSignatures()`

packages/wallet/src/signer.ts
pkieltyka: btw, JsonRpcProvider is also of type Provider, so if something requires Provider, it will still be accepted.

packages/provider/src/wallet-request-handler.ts
pkieltyka: or if we assume its a unique format, perhaps we can leave a comment here, to explain? would be helpful to future reader

packages/wallet/tests/wallet.spec.ts
Agusx1211: `aggregate` supports any kind of signature, it doesn't matter if it's a transaction signature or a message signature. I think renaming it to `joinSignatures` works

packages/provider/src/wallet-request-handler.ts
Agusx1211: Yes, the idea is that this format is not meant to be handled by an RPC node directly, instead it only works if the transaction is later relayed using this Sequence handler

PR #90, opened 2021-01-26
Use indexed image hash and config lookup

Agusx1211:
Depends on https://github.com/0xsequence/wallet-contracts/pull/107

PR #95, opened 2021-01-29
Use indexed wallet for config lookup - (alt)

Agusx1211:
Replaces #90 

- Uses indexed wallet to lookup event
- Uses `publishInitialSigners` for counter-factual config publish

Depends on: https://github.com/0xsequence/wallet-contracts/pull/109

### DIFF
--- a/packages/wallet/src/account.ts
+++ b/packages/wallet/src/account.ts
@@ -321,7 +321,7 @@ export class Account extends Signer {
     const logBlockHeight = (await authContract.lastWalletUpdate(this.address)).toNumber()
     const filter = authContract.filters.RequiredConfig(this.address)
     const lastLog = await findLatestLog(authWallet.provider, { ...filter, fromBlock: logBlockHeight, toBlock: logBlockHeight !== 0 ? logBlockHeight : 'latest'})
-    if (lastLog === undefined) { console.log("LOG NOT FOUND"); return undefined }
+    if (lastLog === undefined) { console.warn("publishConfig: wallet config last log not found"); return undefined }
     const event = authContract.interface.decodeEventLog('RequiredConfig', lastLog.data, lastLog.topics)
 
     const signers = ethers.utils.defaultAbiCoder.decode(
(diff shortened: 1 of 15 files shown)


packages/wallet/src/account.ts
```diff
@@ -262,8 +262,8 @@ export class Account extends Signer {
 
   // publishConfig will publish the wallet config to the network via the relayer. Publishing
   // the config will also store the entire object of signers.
-  publishConfig(): Promise<TransactionResponse> {
-    return this.authWallet().wallet.publishConfig()
+  publishConfig(indexed?: boolean): Promise<TransactionResponse> {
```
pkieltyka, line 265: I noticed the argument name is `index` in the updateConfig method, lets keep these consistent. I think the name `indexed` is a bit more clear, so shall we update all to be consistent?

packages/wallet/src/account.ts
```diff
@@ -306,33 +306,24 @@ export class Account extends Signer {
...
-      // The wallet it's using the counter-factual configuration, get the init config from the authChain
-      const filter = authContract.filters.RequiredConfig(address)
-      const lastLog = await findLatestLog(authWallet.provider, { ...filter, fromBlock: 0, toBlock: 'latest'})
-      if (lastLog === undefined) return undefined
-      event = authContract.interface.decodeEventLog('RequiredConfig', lastLog.data, lastLog.topics)
     }
 
+
+    // Get last known configuration
+    const logBlockHeight = (await authContract.lastWalletUpdate(this.address)).toNumber()
```
pkieltyka, line 321: in case this has not been set, will the contract return `0` value? just so we can test properly for the default

packages/wallet/src/account.ts
```diff
@@ -306,33 +306,24 @@ export class Account extends Signer {
...
+    const filter = authContract.filters.RequiredConfig(this.address)
+    const lastLog = await findLatestLog(authWallet.provider, { ...filter, fromBlock: logBlockHeight, toBlock: logBlockHeight !== 0 ? logBlockHeight : 'latest'})
+    if (lastLog === undefined) { console.log("LOG NOT FOUND"); return undefined }
```
pkieltyka, line 324: lets update this console log to.. `console.warn("publishConfig: wallet config last log not found")`

pkieltyka:
+ we should update the `sequenceContext` in network/context.ts

packages/wallet/src/account.ts
Agusx1211: Yes, returns 0 and defaults to what we had previously

PR #102, opened 2021-02-04
Implement sessions

Agusx1211:
- Move `currentConfig` to SequenceUtilsFinder
- Implement `findLastWalletOfInitialSigner`
- Create Session class
- Manage opening sessions using reference signer
- Manage JWT Api tokens
- Handle expired JWT tokens


packages/api/package.json
```diff
@@ -12,7 +12,9 @@
     "test": "echo",
     "typecheck": "tsc --noEmit"
   },
-  "dependencies": {},
+  "dependencies": {
+    "cross-fetch": "^3.0.6"
```
pkieltyka, line 16: good update here, this is helpful to support node clients.

packages/0xsequence/package.json
```diff
@@ -46,6 +46,7 @@
   "devDependencies": {
     "@babel/plugin-transform-runtime": "^7.12.1",
     "babel-loader": "^8.2.1",
+    "hardhat": "2.0.8",
```
pkieltyka, line 49: Do we need this here as we do define it in /package.json, and then all pkgs share node_modules. It's not a big deal if we have it here too, but just wondering if it wasn't working for you previously? or why else to add it?

packages/api/src/index.ts
```diff
@@ -22,7 +23,7 @@ export class ArcadeumAPIClient extends BaseArcadeumAPI {
       // before the request is made
       init!.headers = { ...init!.headers, ...headers }
   
-      window.fetch(input, init).then(resp => {
+      fetch(input, init).then(resp => {
```
pkieltyka, line 26: thank you

packages/auth/src/session.ts
```diff
@@ -0,0 +1,302 @@
...
+      session.setConfig(newConfig)
+      session.setAccount(new Account({
+        initialConfig: newConfig,
+        networks: networks,
+        context: context
+      }, ...signers.map((s) => s.signer)))
+
+      // Fire JWT requests again, but with new config
+      // MAYBE This is not neccesary, we can rely on the first request?
+      networks.forEach((n) => session.scheduleAuth(n))
```
pkieltyka, line 143: yes we can rely on the first one. I think for ethauth to get a jwt token -- we can just submit a partial signing from the login key or any key in the wallet config prior to this new session. Is it possible with for us to do partial signature validation with 1271? the idea is ethauth is used on both client and server as a signing of eip712 payload and then validation on server. On ethauth.js (v0.4.3), we don't need to validate the signature upon signing, we can just do a partial signing of the eip712 and fire it off to the server for exchange for a jwt. We have go-ethauth on server side which can also sign or validate a eip712 "claim" / "proof", and here we could validate a partial signature and in the response of the proof we could return the signers which have successfully validated the 712 ethauth claim. Then on the jwt handler where we take an ethauth proof and exchange for a jwt, we can check the signer list and see if the threshold is at least up to one gold key, at which point, we will issue the jwt.

pkieltyka (commented):
Looking good -- I will spend more time with it over the weekend where I'll pull the branches and try it in the application code so I can get even better feel for it. It's a huge step forward, thank you. We'll also smooth out a few things in ethauth so we can support partial signing and avoid having to do extra checks when unnecessary, which will speed up exec too

packages/auth/src/session.ts
pkieltyka: I have a better solution to removing this secondary auth schedule here: https://github.com/0xsequence/sequence.js/pull/102/commits/a319d3795177de55e1c2cf5e30e814bd6e7571e2#diff-da2446c8cacd86376976dc5132ed9d5de20688e372f61f80d15f02f3061261fbR129-R143

+ makes a number of other improvements and performance updates as a result

Agusx1211: The issue is not that we are over-signing with signers that later become invalid, the issue is that Sequence signature encoding and wallet configuration is intertwined, so when you update the configuration of the wallet, old signatures that should still be valid become invalid because the encoding changes, for example:

```
Original configuration:
Signer A - Signer B - Signer C

Signed Transaction 1:
Signature A - Signature B - Address C

Change configuration to:
Signer A - Signer B - Signer D

Now the signed transaction 1 is no longer valid, because it should be encoded as:
Signature A - Signature B - Address D
```

There are two man issues to solve it:

1 - EIP1271 signatures don't have any way of validating "partial signatures".
2 - The JWT service doesn't have a way of retrieving the contents of the latest configuration of the wallet, it only has the "hash" of the configuration and thus needs the full configuration in some way to validate a signature.

This being a race condition I think we can find more simple solutions, one could be waiting for the JWT token before updating the wallet (this is what happens most of the time anyway, the second "try" is more to catch a rare race condition when the wallet updates faster than the JWT generation), this should work 100% of the time and it's clean, but it would increase the time it takes to open the session.

Another option is to send a specific "block" that should be used to check the signature, then the JWT server can validate the signature using the wallet at the state of the given block, effectively also enforcing order but without blocking. There are two details we need to keep in mind before going this route:

1) The JWT server would need to have some limit on how old the requested block can be, otherwise it would open a secuirty issue where an attacker can get a JWT token using an old configuration.
2) The JWT client would need to get the current block of the network before sending the JWT request, the server would also need to know the current block of the network to check the age of the block, those two things combined could make this process more slow than just waiting for the original JWT as proposed before.

packages/0xsequence/package.json
Agusx1211: It will work if other packages also use hardhat as a dependency, but technically this package also does, and some tools give warnings if it's not properly defined

PR #114, opened 2021-03-12
Fix loading config from intermediary state on require utils

Agusx1211:
- Add tests for Sequence utils finder
- Fix loading config from non-auth chain on counterfactual states
- Fix loading config from non-auth chain on intermediary states


acrylix (approved):
tested on mainnet, works!
this was a critical fix, thanks @Agusx1211

PR #119, opened 2021-03-22
session: add support for public address as signers for session.open

acrylix:
.


pkieltyka (approved):
LGTM, but lets ask @Agusx1211 to review tomm too

PR #125, opened 2021-04-07
improve openWindow state options handling + typings

acrylix:
added extra state options in openWindow to distinguish between jsonRpcRequest and regular wallet open 
this improves popup loading performance from dapps

instead of always loading root path (/wallet) and corresponding HTTP requests when a popup is open, the new states allows the wallet to distinguish openWindow intent and display an intermediary loading route, then proceed straight to the corresponding jsonRpcRequest handler routes

related: https://github.com/horizon-games/issue-tracker/issues/3936

### DIFF
--- a/packages/provider/src/transports/base-provider-transport.ts
+++ b/packages/provider/src/transports/base-provider-transport.ts
@@ -4,7 +4,7 @@ import {
   ProviderTransport, ProviderMessage, ProviderMessageRequest,
   ProviderMessageType, ProviderMessageEvent, ProviderMessageResponse,
   ProviderMessageResponseCallback, ProviderMessageTransport,
-  WalletSession, ConnectionState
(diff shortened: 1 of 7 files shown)


packages/provider/src/transports/base-provider-transport.ts
```diff
@@ -49,7 +49,7 @@ export abstract class BaseProviderTransport implements ProviderTransport {
     throw new Error('abstract method')
   }
 
-  openWallet(path?: string, state?: any, defaultNetworkId?: string | number) {
+  openWallet(path?: string, state?: OpenWalletState, defaultNetworkId?: string | number) {
```
pkieltyka, line 52: all looks good, perhaps we should rename the variable name from `state` to `intent` and then `OpenWalletState` -> `OpenWalletIntent` ?

PR #126, opened 2021-04-08
Update multicall documentation

epg323:
The current documentation did not work for me, probably due to some changes in the codebase. I had to import differently to get it to work.


Agusx1211:
Good catch, that was some old how-to, thanks!

epg323:
My pleasure 😎

PR #132, opened 2021-04-20
Implement knownImageHashes index and refactor sequence utils finder

Agusx1211:
image

### DIFF
--- a/packages/config/src/finder/sequence-utils-finder.ts
+++ b/packages/config/src/finder/sequence-utils-finder.ts
@@ -22,102 +22,20 @@ export class SequenceUtilsFinder implements ConfigFinder {
 
     if (requireIndex && ignoreIndex) throw Error('Can\'t ignore index and require index')
 
+    const chainIdPromise = provider.getNetwork()
     const knownConfigs = args.knownConfigs ? args.knownConfigs : []
 
-    const chainId = (await provider.getNetwork()).chainId
+    // Get imageHash of wallet
+    const { imageHash, config } = await this.findCurrentImageHash(context, provider, address, knownConfigs)
+    if (imageHash === undefined) return { config: undefined }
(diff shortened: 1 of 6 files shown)


packages/config/src/finder/sequence-utils-finder.ts
```diff
@@ -22,122 +22,127 @@ export class SequenceUtilsFinder implements ConfigFinder {
...
+    requireIndex?: boolean
+  }): Promise<{ wallet: string | undefined }> => {
+    const { signer, context, ignoreIndex, requireIndex } = args
+    if (requireIndex && ignoreIndex) throw Error('Can\'t ignore index and require index')
 
     const authContract = new Contract(context.sequenceUtils!, walletContracts.sequenceUtils.abi, this.authProvider)
+    const logBlockHeight = ignoreIndex ? 0 : (await authContract.lastSignerUpdate(signer)).toNumber()
+    if (requireIndex && logBlockHeight === 0) return { wallet: undefined }
+    const filter = authContract.filters.RequiredSigner(null, signer)
+    const lastLog = await this.findLatestLog(this.authProvider, { ...filter, fromBlock: logBlockHeight, toBlock: logBlockHeight !== 0 ? logBlockHeight : 'latest'})
```
pkieltyka, line 55: can we add a `logger.warn()` in here with new `@0xsequence/utils` when logBlockHeight is 0, this way, we know it's doing a very long/expensive log query, and at least we can log this, then we can alert ourselves in future. Very helpful to know when these scenarios are triggered, and we should warn as we should do all we can to prevent it. I assume with latest code, it won't happen anymore, but its a good thing to have this "observability" in the wild

pkieltyka (approved):
LGTM! my only feedback was a comment and a logger line

PR #139, opened 2021-05-06
prompt connect

acrylix:


packages/provider/src/transports/base-provider-transport.ts
```diff
@@ -147,9 +156,13 @@ export abstract class BaseProviderTransport implements ProviderTransport {
       return
     }
 
+    if (message.type === ProviderMessageType.AUTHORIZED) {
```
pkieltyka, line 159: we should remove "AUTHORIZED" event, and just use same event/signal as CONNECT for this data. Otherwise it creates multiple codepaths and async issues. I've experienced a bug when trying to connect + auth from cold start and it never asks me to authorize -- or does it perhaps authorize for me automatically?

packages/provider/src/transports/base-provider-transport.ts
```diff
@@ -345,9 +357,27 @@ export abstract class BaseProviderTransport implements ProviderTransport {
       })
     })
 
-    return Promise.race<WalletSession>([
-      connect,
-      closeWallet
+    return Promise.race<WalletSession>([connect, closeWallet])
+  }
+
+  waitUntilAuthorized = async (): Promise<ConnectDetails> => {
```
pkieltyka, line 363: lets remove and just have waitUntilConnected(). If a dapp makes a request to `connect({ requestAuthorization })`, then we need to respond with connect and inside provide the authorization

packages/provider/src/transports/base-wallet-transport.ts
```diff
@@ -126,6 +126,16 @@ export abstract class BaseWalletTransport implements WalletTransport {
     })
   }
 
+  notifyAuthorized(connectDetails: ConnectDetails) {
```
pkieltyka, line 129: lets remove

packages/provider/src/transports/mux-transport/mux-message-provider.ts
```diff
@@ -147,4 +147,11 @@ export class MuxMessageProvider implements ProviderTransport {
     throw new Error('impossible state, wallet must be opened first')
   }
 
+  waitUntilAuthorized = async(): Promise<ConnectDetails> => {
```
pkieltyka, line 150: remove

packages/provider/src/transports/window-transport/window-message-handler.ts
```diff
@@ -30,23 +33,41 @@ export class WindowMessageHandler extends BaseWalletTransport {
...
-    const jsonRpcRequest = location.searchParams.get('jsonRpcRequest')
-
-    if (options?.loadingPath && !!jsonRpcRequest) {
-      window.history.replaceState({}, document.title, options.loadingPath)
+    const defaultNetwork = params['net']
+
+    const toBool = (s: string) => s === 'true'
+
+    if (toBool(params['requestAuthorization'])) {
+      this.walletRequestHandler
```
pkieltyka, line 52: the WindowMessageHandler should not be calling `this.walletRequestHandler` ever

`this.walletRequestHandler` should be called from the base class of `BaseWalletTransport` so other implementations will follow the same code-path. Ie. proxy-transport and soon post-message-transport (for chrome extension)

PR #140, opened 2021-05-06
auth: piggyback on already pending JWT and signing requests

attente:
If a session is already in the process of obtaining a JWT or authentication signature, reuse the existing requests to avoid spamming the signers and servers with multiple identical requests.

https://github.com/horizon-games/issue-tracker/issues/4689

Co-authored-by: Ari Lotter <arilotter@gmail.com>
Co-authored-by: Michael Yu <1619025+acrylix@users.noreply.github.com>


packages/auth/src/session.ts
```diff
@@ -67,109 +90,192 @@ export class Session implements SessionDump {
...
+    this.proofStrings.set(url, proofString)
 
-    // TODO: ethauth.js v0.4.4:
-    // const proofString = await ethAuth.encodeProof(proof, { skipValidation: true })
+    return proofString
+  }
 
-    const url = (await network).sequenceApiUrl
-    if (!url) return
+  private async isProofStringValid(proofString: string, network: NetworkConfig): Promise<boolean> {
```
acrylix, line 240: make sure network here is using auth chain network for signature validation

packages/api/src/index.ts
```diff
@@ -1,35 +1,68 @@
...
+    let url: string
+    switch (typeof input) {
+      case 'string':
+        url = input
+        break
+      case 'object':
+        url = input.url
+        break
+    }
+    const isGetAuthToken = new URL(url).pathname.endsWith('/GetAuthToken')
```
arilotter, line 40: I'm not sure if this `_fetch` could ever get called with user-supplied URLs, but if it could, there's no guarantee that the URL passed to a fetch is a "full" URL, it might just be a path "/test" or something else.

One solution is to wrap in try/catch.

packages/auth/src/session.ts
```diff
@@ -67,109 +90,192 @@ export class Session implements SessionDump {
...
   }
 
-  async performAuthRequest(net: NetworkConfig | number): Promise<void> {
-    const network = this.getNetwork(net)
+  private getProofString(url: string): ProofStringPromise {
+    // check if we already have or are waiting for a proof string
+    if (this.proofStrings.has(url)) {
+      const proofString = this.proofStrings.get(url)!
+
+      if (this.now() < proofString.expiration) {
```
arilotter, line 205: What happens if our clock is out of sync with the server, so the server will reject our proof string, but it looks good to us?

packages/auth/src/session.ts
```diff
@@ -67,109 +90,192 @@ export class Session implements SessionDump {
...
     const proof = new Proof({
       address: this.account.address
     })
-  
+    proof.claims.app = this.name
     proof.setIssuedAtNow()
     proof.setExpiryIn(this.expiration)
 
+    const ethAuth = new ETHAuth()
+    // TODO: is authWallet up-to-date?
```
arilotter, line 221: is it?

packages/auth/tests/session.spec.ts
```diff
@@ -120,7 +162,7 @@ describe('Wallet integration', function () {
   it("Should dump and load a session", async () => {
     const referenceSigner = ethers.Wallet.createRandom()
 
-    const dump = (await Session.open({
+    const dump = await (await Session.open({
```
arilotter, line 165: `await (await (` is kinda hard to parse, it took me a sec to realize i was awaiting the result of `dump`. Maybe this would be easier to read as `await Session.open({}).then(s => s.dump())`

packages/auth/tests/session.spec.ts
```diff
@@ -528,21 +570,21 @@ describe('Wallet integration', function () {
         }
       })
 
-      expect(session.authPromises.length).to.equal(1)
-      await Promise.all(session.authPromises.map((p) => p.promise))
+      expect((session as any)._allInitialAuthRequests.length).to.equal(networksWithApi.length)
```
arilotter, line 573: is this `as any` so you can pull out the private property? seems stinky to be testing an internal property of the system

packages/auth/tests/session.spec.ts
```diff
@@ -565,15 +607,16 @@ describe('Wallet integration', function () {
           name: "Test"
         }
       })
-  
-      expect(session.authPromises.length).to.equal(2)
-      await Promise.all(session.authPromises.map((p) => p.promise))
 
-      expect(totalCount).to.equal(2)
-      expect(recoverCount["error"]).to.equal(1)
+      expect((session as any)._allInitialAuthRequests.length).to.equal(networksWithApi.length)
```
arilotter, line 611: again, don't like testing to read a private property. super brittle. this'll also break if the session's internal impl changes at all

packages/auth/tests/session.spec.ts
```diff
@@ -757,15 +800,19 @@ describe('Wallet integration', function () {
...
+
+      const totalCountBefore = totalCount
 
+      // This should use the already existing JWT
       const api = await session.getAPI(networksWithApi[0])
 
-      expect(totalCount).to.equal(2)
-      expect(recoverCount["error"]).to.equal(1)
+      expect(totalCount).to.equal(totalCountBefore)
+      // TODO: can't reproduce
```
arilotter, line 811: should this be resolved before PR is merged?

arilotter (changes requested):

packages/api/src/index.ts
attente: In the generated bindings, it always uses fully resolved URLs: https://github.com/0xsequence/sequence.js/blob/1a3c9f43b38b94b87af1a47bd537ea9ff92f7479/packages/api/src/api.gen.ts#L812, but I'll make this more robust so it doesn't have to assume that.

packages/auth/src/session.ts
attente: Pretty much all `GetAuthToken` requests to the API will fail in that case, but that's no different from the current behaviour.

packages/auth/src/session.ts
attente: nope

packages/auth/src/session.ts
arilotter: Is there something clever we can do about checking our assumed system time vs the timestamp returned by some trusted service to avoid the GetAuthToken spam in these cases?

packages/api/src/index.ts
```diff
@@ -1,35 +1,73 @@
...
+      isGetAuthToken = url.includes('/GetAuthToken')
+    }
+
+    try {
+      const resp = await fetch(input, init)
+
+      if (resp.status >= 400 && resp.status < 600) {
+        if (retry && this.refreshJWT && !isGetAuthToken) {
+          // refresh token...
+          this.jwtAuth = await this.refreshJWT()
```
pkieltyka, line 53: How about we remove this auto-refreshing of JWTs..? the reason is, ethauth, and the jwt token made from it will only be valid for the duration the token is allowed to be valid for.. 

I feel removing the auto-refresh will simplify the code? But I'm trying to understand why we need it..?

PR #147, opened 2021-05-11
Strip 0x from subdigest hash

corbanbrook:


packages/transactions/src/utils.ts
```diff
@@ -24,7 +24,7 @@ export function digestOfTransactions(...txs: Transaction[]): string {
 }
 
 export function subdigestOfTransactions(address: string, chainId: BigNumberish, ...txs: Transaction[]): string {
-  return subDigestOf(address, chainId, digestOfTransactions(...txs))
+  return subDigestOf(address, chainId, digestOfTransactions(...txs)).replace(/^0x/, '') // TODO Backend should store the hash with 0x prefix
```
pkieltyka, line 27: how come we have to do this..?

PR #151, opened 2021-05-20
auth: JWT and signing request piggybacking again

attente:
The regression turned out to be caused by a hard-coded constant gas limit for config updates that ends up reverting if the config accumulates too many signers (~50 or so).

### DIFF
--- a/packages/wallet/src/wallet.ts
+++ b/packages/wallet/src/wallet.ts
@@ -521,7 +521,7 @@ export class Wallet extends Signer {
 
     const [txs, n] = await Promise.all([
       this.buildUpdateConfigTransaction(config, publish, indexed),
-      nonce ? nonce : await this.getNonce()
+      nonce ?? this.getNonce()
     ])
 
     return [
--- a/packages/wallet/src/wallet.ts
+++ b/packages/wallet/src/wallet.ts
@@ -554,9 +554,11 @@ export class Wallet extends Signer {
 
     const walletInterface = new Interface(walletContracts.mainModule.abi)
 
-    // 131072 gas, enough for both calls
-    // and a power of two to keep the gas cost of data low
-    const gasLimit = ethers.constants.Two.pow(17)
+    // empirically, this seems to work for the tests:
+    // const gasLimit = 100000 + 1800 * config.signers.length
+    //
+    // but we're going to play it safe with this instead:
+    const gasLimit = 2 * (100000 + 1800 * config.signers.length)
 
     const preTransaction = isUpgradable ? [] : [{
       delegateCall: false,
--- a/packages/wallet/src/wallet.ts
+++ b/packages/wallet/src/wallet.ts
@@ -594,7 +594,7 @@ export class Wallet extends Signer {
 
     return [{
       delegateCall: false,
-      revertOnError: false,
+      revertOnError: true,
       gasLimit: gasLimit,
       to: this.address,
       value: ethers.constants.Zero,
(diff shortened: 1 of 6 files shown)


attente:
Notes:
- The revert commit is the same as the original PR, it has not changed at all
- We tested this by adding ~50 signers to the wallet config which broke as observed
- This branch has also been tested manually to confirm that the problem was fixed
- The new gas limit calculation is super conservative, I made some measurements empirically and then scaled up by a safety factor of 2. In practice it seems to lead to a 3x actual gas usage, see https://explorer-mainnet.maticvigil.com/tx/0xefbdbec551b7aaf17b667a36707a790fe4f3bb112f75a82f86b955b75beb3cbb/internal-transactions and https://explorer-mainnet.maticvigil.com/tx/0x3ff8da827b674ca13c71820d90aca6f7ad1ee02a5b71db3482df99f139ef857a/internal-transactions
- @Agusx1211 We had to enable `revertOnError` on the bundled config update transaction so that the tests could detect the failure case: https://github.com/0xsequence/sequence.js/pull/151/commits/4ac7b72ce51d3e276a8838a2b6054c2023ac6829. I think that makes sense since I think the bundle of transactions should be done as an atomic operation, but if there's a good reason it shouldn't be, please let me know

acrylix (approved):

packages/wallet/src/wallet.ts
```diff
@@ -521,7 +521,7 @@ export class Wallet extends Signer {
 
     const [txs, n] = await Promise.all([
       this.buildUpdateConfigTransaction(config, publish, indexed),
-      nonce ? nonce : await this.getNonce()
+      nonce ?? this.getNonce()
```
attente, line 524: BTW, this actually changes behaviour, so please let me know if this is OK. Before, if `nonce === 0`, this would make an async call to `this.getNonce()`. Now, it actually return 0 directly if `nonce === 0`.

PhABC: This may be better actually, but I'll let @Agusx1211 confirm as well.

packages/wallet/tests/wallet.spec.ts
```diff
@@ -317,13 +317,55 @@ describe('Wallet integration', function () {
...
+            }
+
+            let tx = (await wallet.updateConfig(newConfig, undefined, true))[1]
+            let receipt = await tx.wait()
+            expect(receipt.status).to.equal(1)
+            const gasLimit1 = tx.gasLimit
+
+            // next, we try 100 signers
+
+            while (signers.length < 100) {
```
PhABC, line 352: Can we test 1000 signers and remove the 2x factor when calculating the gas limit above? I'd rather have a tighter gas limit estimate.

packages/wallet/src/wallet.ts
```diff
@@ -594,7 +596,7 @@ export class Wallet extends Signer {
 
     return [{
       delegateCall: false,
-      revertOnError: false,
+      revertOnError: true,
```
PhABC, line 599: @Agusx1211 to confirm if this is OK

packages/wallet/tests/wallet.spec.ts
pkieltyka: or use 1.25x factor if we feel we need the safety? as well, if we fail-fast, and catch errors, then we'll know at least when we hit issues with gas. In the future, we should always remember when something breaks to always just check the relayer wallet

pkieltyka:
I'm going to merge this + release v0.21 so we can try with dev. If there is any follow up work from comments above, we can apply a PR on top of that -- just to keep things moving forward

PR #152, opened 2021-05-20
wallet: config updates must not be revertOnError

attente:


packages/wallet/src/wallet.ts
```diff
@@ -596,7 +596,7 @@ export class Wallet extends Signer {
 
     return [{
       delegateCall: false,
-      revertOnError: true,
+      revertOnError: false,
```
attente, line 599: This is the main change.

Agusx1211 (approved):
lgtm

PR #153, opened 2021-05-21
Throwing exception from wait for failed statuses

corbanbrook:


Agusx1211 (approved):

packages/relayer/src/rpc-relayer/index.ts
```diff
@@ -170,8 +165,13 @@ export class RpcRelayer extends BaseRelayer implements Relayer {
     return this.wait(signedTxs)
   }
 
-  async wait(metaTxnHash: string | SignedTransactions, wait: number = 500): Promise<TransactionResponse> {
+  async wait(metaTxnHash: string | SignedTransactions, wait: number = 1000): Promise<TransactionResponse> {
     const { receipt } = await this.waitReceipt(metaTxnHash, wait)
+
+    if (!receipt.txnReceipt || receipt.status === 'FAILED' || receipt.status === 'DROPPED') {
```
PhABC, line 171: What about the other "error" like statuses?

```go
ETHTxnStatus_UNKNOWN             ETHTxnStatus = 0
ETHTxnStatus_PARTIALLY_FAILED    ETHTxnStatus = 2
ETHTxnStatus_REVERTED            ETHTxnStatus = 4
```

The latter two are for when the tx actually reverts (or a tx within a batch)

PR #157, opened 2021-06-03
Relayer options

Agusx1211:
- Add parameters for relayer options
- Split ProviderRelayer and LocalRelayer
- Allow for custom `creationGasLimit`

This change is required to support more complex relayer options without breaking interfaces


acrylix (approved):

pkieltyka:
Looking good -- the goal is to allow any provider to act as a relayer? ie. this way a funded metamask wallet could relay txns..? 

I wonder too what is the best interface we should have to allow third-party service-oriented relayers..? Services could implement our RpcRelayer interface, which is a pretty simple set of 4 endpoints on /rpc/Relayer, but maybe we can use standard Ethereum JSON-RPC too

corbanbrook (approved):

PR #158, opened 2021-06-03
Integrate relayer options in networks

Agusx1211:


packages/network/src/utils.ts
```diff
@@ -75,7 +75,7 @@ export const isValidNetworkConfig = (networkConfig: NetworkConfig | NetworkConfi
       return false
     }
     if (!skipRelayerCheck) {
-      if ((!c.relayerUrl || c.relayerUrl === '') && !c.relayer) {
+      if (!c.relayer || (typeof c.relayer === 'string' && c.relayer === '')) {
```
pkieltyka, line 78: seems the c.relayer will never be of type 'string' ?

Agusx1211: Good catch!

packages/utils/src/big-number.ts
```diff
@@ -0,0 +1,14 @@
+import { BigNumber, BigNumberish } from "ethers";
```
pkieltyka, line 1: nice! good addition to utils pkg

PR #161, opened 2021-06-08
Fix invalid ETHauthproof when session expired

PhABC:
Closes https://github.com/horizon-games/issue-tracker/issues/4953

### DIFF
--- a/packages/auth/src/session.ts
+++ b/packages/auth/src/session.ts
@@ -208,14 +208,26 @@ export class Session {
     proof.setExpiryIn(this.expiration)
 
     const ethAuth = new ETHAuth()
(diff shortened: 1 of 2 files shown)


packages/auth/src/session.ts
```diff
@@ -208,22 +208,33 @@ export class Session {
     proof.setExpiryIn(this.expiration)
 
     const ethAuth = new ETHAuth()
-    // TODO: is authWallet up-to-date?
```
PhABC, line 211: Narrator: "It was not."

acrylix (approved):

Agusx1211 (approved):
LGTM!

PR #163, opened 2021-06-08
update relayer bindings

attente:


packages/relayer/src/rpc-relayer/index.ts
```diff
@@ -96,55 +96,35 @@ export class RpcRelayer extends BaseRelayer implements Relayer {
     return prevNonce === undefined ? modTxns : appendNonce(modTxns, prevNonce)
   }
 
-  async gasRefundOptions(config: WalletConfig, context: WalletContext, ...transactions: Transaction[]): Promise<Transaction[][]> {
-    // chaind only supports refunds on a single token
-    // TODO: Add compatiblity for different refund options
-    const tokenFee = await this.service.tokenFee()
+  async gasRefundOptions(config: WalletConfig, context: WalletContext, ...transactions: Transaction[]): Promise<FeeOption[]> {
+    const feeTokens = await this.service.feeTokens()
```
pkieltyka, line 100: in the future, we should memoize this so we don't need to ask each time for a network's fee token options, but for first version this is fine. Perhaps you could add a TODO comment above?

pkieltyka (approved):

PR #167, opened 2021-06-15
wallet: remove duplicate contract wallet validation

pkieltyka:


packages/wallet/src/validate.ts
```diff
@@ -20,18 +20,16 @@ export async function isValidSignature(
     isValidEthSignSignature(address, digest, sig)
   ) return true
 
-  const wallets = await Promise.all([
-    isValidContractWalletSignature(address, digest, sig, provider),
-    isValidSequenceDeployedWalletSignature(address, digest, sig, provider)
```
PhABC, line 25: Why were we checking deployed wallet signature but now we are checking undeployed wallet signature?

pkieltyka: We're still checking a deployed wallet signature (just line below, at `const erc1271Check = await isValidContractWalletSignature(address, digest, sig, provider)`) but instead of a promise with multiple calls which do the same thing, we're just doing a single erc1271 check. I think this mistake dates back to a subdigest discrependency where we were computing an extra subdigest on sequence and so we needed separate functions, but now a single erc1271 check is enough

PR #171, opened 2021-06-23
Implement ProviderRelayer wait and test it

Agusx1211:


packages/relayer/tests/provider-relayer.spec.ts
```diff
@@ -0,0 +1,445 @@
+import { deployWalletContext } from './utils/deploy-wallet-context'
```
pkieltyka, line 1: is it not possible to `import { deployWalletContext } from '../../0xsequence/utils/deploy-wallet-context'` .. (or whatever the path is, but I'm pretty sure we have this function implemented in wither packages/wallets or packages/0xsequence

Agusx1211: Yes we have that duplicated code everywhere, but I didn't want to mix changing the pattern in this PR, I agree it's better to move it to the utils package and clean up all other tests, but I would leave that for a different PR.

packages/relayer/src/provider-relayer.ts
```diff
@@ -1,16 +1,24 @@
...
 import { BaseRelayer, BaseRelayerOptions } from './base-relayer'
 import { FeeOption, Relayer } from '.'
+import { Optionals, Mask } from '@0xsequence/utils'
 
 const DEFAULT_GAS_LIMIT = ethers.BigNumber.from(800000)
 
 export interface ProviderRelayerOptions extends BaseRelayerOptions {
-  provider: Provider
+  provider: Provider,
+  waitPoolRate?: number,
```
pkieltyka, line 15: whats `waitPoolRate`?

packages/relayer/src/provider-relayer.ts
```diff
@@ -1,16 +1,24 @@
...
+  deltaBlocksLog?: number
```
pkieltyka, line 16: whats `deltaBlocksLog`? doesnt seem we're using it either anywhere else in the code

packages/relayer/src/provider-relayer.ts
PhABC: I think it's meant to be `waitPollRate`, i.e. the polling rate.

packages/relayer/src/provider-relayer.ts
```diff
@@ -91,7 +104,61 @@ export abstract class ProviderRelayer extends BaseRelayer implements Relayer {
...
+    // get all nonce changes and look for metaTxnIds in between logs
+    const timeoutTime = new Date().getTime() + timeout
+    let lastBlock: number = 0
+
+    const normalMetaTxnId = metaTxnId.replace('0x', '')
+
+    while (new Date().getTime() < timeoutTime) {
+      const block = await this.provider.getBlockNumber()
+      const logs = await this.provider.getLogs({
+        fromBlock: Math.max(0, lastBlock - 12),
```
PhABC: Should be using the `deltaBlocksLog` variable, not hardcoded 12

packages/relayer/src/provider-relayer.ts
```diff
@@ -91,7 +104,61 @@ export abstract class ProviderRelayer extends BaseRelayer implements Relayer {
...
+      // If found return that
+      if (found) {
+        return {
+          ...found,
+          ...await this.provider.getTransaction(found.transactionHash)
+        }
+      }
+
+      // Otherwise wait and try again
+      await new Promise(r => setTimeout(r, 1000))
```
PhABC: Should be using the waitPollRate variable here, not hardcoding 1000

packages/relayer/src/provider-relayer.ts
```diff
@@ -91,7 +104,61 @@ export abstract class ProviderRelayer extends BaseRelayer implements Relayer {
...
+
+    const normalMetaTxnId = metaTxnId.replace('0x', '')
+
+    while (new Date().getTime() < timeoutTime) {
+      const block = await this.provider.getBlockNumber()
+      const logs = await this.provider.getLogs({
+        fromBlock: Math.max(0, lastBlock - 12),
+        toBlock: block,
+        // Nonce change event topic
+        topics: ['0x1f180c27086c7a39ea2a7b25239d1ab92348f07ca7bb59d1438fcf527568f881']
```
PhABC, line 138: Can we get the topics from the ABI by anychance? Would be a bit cleaner.

Also, can we filter by wallet address? I.e. the contract address that emits this event? Would probably return less logs.

Agusx1211: We could filter by wallet address, but that's not currently part of the relayer interface, we would need to do a bigger change and add it to the interface

packages/relayer/src/provider-relayer.ts
Agusx1211: Yep, it was an oversight, I just fixed it :sweat_smile:

packages/relayer/src/provider-relayer.ts
Agusx1211: That should be used instead of the hardcoded values, I fixed it, thanks!

packages/relayer/src/provider-relayer.ts
PhABC: Right, we are only passing the metatxnID, I see. Lets keep it for the future, if we see polling without such filter is too strenuous (e.g. if we have 1000s of wallets doing txs at the same time).

PR #172, opened 2021-07-05
Gas estimation module

Agusx1211:
- Estimate gas usage using `eth_call` and overwrites
- Use docker geth instance for testing
- Compute gas usage of transactions in batch by measuring delta between partial estimations
- Support nested Sequence wallets up to a depth of 1 (assumes 2/3 config for the child sequence wallets)
- Account for calldata and base cost

### DIFF
--- a/packages/estimator/src/overwriter-sequence-estimator.ts
+++ b/packages/estimator/src/overwriter-sequence-estimator.ts
@@ -1,22 +1,92 @@
 import { WalletContext } from '@0xsequence/network'
-import { WalletConfig, addressOf } from '@0xsequence/config'
-import { Transaction } from '@0xsequence/transactions'
+import { WalletConfig, addressOf, encodeSignature } from '@0xsequence/config'
+import { readSequenceNonce, Transaction } from '@0xsequence/transactions'
 import { OverwriterEstimator } from './overwriter-estimator'
+import { Interface } from 'ethers/lib/utils'
+import { walletContracts } from '@0xsequence/abi'
+import { ethers } from 'ethers'
 
 export class OverwriterSequenceEstimator {
   constructor(public estimator: OverwriterEstimator) {}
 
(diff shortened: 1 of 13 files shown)


packages/estimator/src/overwriter-sequence-estimator.ts
```diff
@@ -0,0 +1,111 @@
...
+      ...s,
+      index: i,
+      isEOA: ethers.utils.arrayify(await this.estimator.provider.getCode(s.address)).length === 0
+    })))
+
+    // Define designated signers
+    let weightSum = 0
+
+    const definedSigners = signers
+      .sort((a, b) => !a.isEOA && b.isEOA ? -1 : 0) // Contract signers sign first
```
PhABC, line 31: I think we should sort by weight as well, so we get the worst case scenario for # of signers required. E.g. imagine we have weights like `[1, 1, 1, 1, 1, 5]` and the threshold is `5`. If the last wallet is the first to be evaluated, then the threshold will be met with a single signature, but possible it could take 5 signers.

attente: Thinking about it, I think the greedy solution doesn't actually work here, although it makes a good first heuristic. Imagine that the first signer of weight 5 is a sequence wallet that requires 100 out of 100 EOA signers. Then the worst case scenario is actually the one dynamic signature here.

PhABC: So you are saying we are doomed

PhABC: The "only" way would be to fetch the config from the logs for every non-EOA signers. Perhaps it wouldn't be too bad because 1. it would simplify the code and 2. it should be rare and most users will mainly have EOAs as signers.

attente: I think this is something like a knapsack problem with a dynamic programming solution if you want to get the worst case scenario for a wallet. But yes, let's just assume the config is always simple for now with just one guard signature being the only non-EOA signer.

PR #173, opened 2021-07-06
fix: support queue higher than batches

nachomazzara:
If the queue's size is high enough to make this true `batchsize < queue - batchsize`.  Subsequent requests are not going to be processed


Agusx1211:
Thanks a lot Nacho!

PR #175, opened 2021-07-08
allow session network config to be updated from default value

acrylix:


packages/0xsequence/tests/browser/wallet-provider/dapp.test.ts
```diff
@@ -102,7 +102,7 @@ export const tests = async () => {
 
   await test('getWalletConfig', async () => {
     const allWalletConfigs = await wallet.getWalletConfig()
-    assert.equal(allWalletConfigs.length, 2, '2 wallet configs (one for each chain)')
+    assert.equal(allWalletConfigs.length, 1, '1 wallet config for default chain')
```
pkieltyka, line 105: the PR makes sense, but how come this assertion goes down from 2 wallet configs to 1?

acrylix: the implementation of `getWalletConfig` actually uses the default chain Id if no value is passed in.
https://github.com/0xsequence/sequence.js/blob/allow-default-network-switch/packages/provider/src/provider.ts#L164
and the default chain value in provider was always set to `undefined` prior to this PR, which resulted in returning all configs with undefined default chain Id. Now that we support default network switching in this PR, default chain Id is explicitly set for each provider, thus affecting the output of getWalletConfig

acrylix: we can adjust this to always return wallet configs for all chains

PR #176, opened 2021-07-16
Extension provider

acrylix:
replaces https://github.com/0xsequence/sequence.js/pull/175


packages/provider/package.json
```diff
@@ -25,10 +25,13 @@
     "@ethersproject/providers": "^5.0.24",
     "@ethersproject/web": "^5.0.14",
     "ethers": "^5.0.32",
-    "eventemitter3": "^4.0.7"
+    "eventemitter3": "^4.0.7",
+    "readable-stream": "^3.6.0"
```
pkieltyka, line 29: lets remove this dep as discussed

packages/provider/src/transports/extension-transport/extension-message-handler.ts
```diff
@@ -0,0 +1,29 @@
...
+  constructor(walletRequestHandler: WalletRequestHandler) {
+    super(walletRequestHandler)
+    this._init = InitState.OK
+  }
+
+  // from within extension execution context
+  register() {
+    this._registered = true
+
+    // TODO: check execution context, throw if not an extension
```
pkieltyka, line 19: good point on the TODO.. we should def check `if ((chrome as any) === undefined) { .. throw..}` or something like it

pkieltyka (commented):
beauty work here! Left two small comments.

The final thing is to support multiple connect's on 0xsequence/provider Wallet as discussed as well

PR #182, opened 2021-08-14
Require fresh signer on new wallet (session)

Agusx1211:


packages/auth/tests/session.spec.ts
```diff
@@ -362,6 +366,40 @@ describe('Wallet integration', function () {
     expect(ogSession.account.address).to.not.equal(session.account.address)
   })
 
+  it("Should fail to open a session by forzing a non-fresh signer", async () => {
```
PhABC, line 369: Forzing? hehe

packages/wallet/src/wallet.ts
```diff
@@ -656,23 +658,41 @@ export class Wallet extends Signer {
...
-        ]
-      )
-    }]
+    const contextRequireFreshSigner = this.context.libs?.requireFreshSigner
+    if (requireFreshSigners.length > 0 && contextRequireFreshSigner === undefined) {
+      throw Error('requireFreshSigners missing library')
+    }
+
+    return [
+      ...(requireFreshSigners.map((signer) => ({
```
PhABC, line 667: Would it be better for the requireFreshSigner to take an array of addresses instead? Right now we need to do n contract calls instead of 1.

packages/auth/tests/session.spec.ts
Agusx1211: This never happened

packages/wallet/src/wallet.ts
Agusx1211: We only use it for the torus signer, and we must likely never use it for anything else. Making it an array would increase the calldata size for our happy path.

PhABC: Oh I see, misread, thought we were iterating over all addresses, not just the `referenceSigner`. All good then

PhABC:
LGTM

PR #188, opened 2021-09-03
Refactor networks

Agusx1211:
This change aims to simplify and make more explicit the definition of networks and defaultChainId, currently sequence.js stores this information both client-side (dApp), on sequence.js (transport) and on the wallet-webapp. This means that any change to the wallet networks configurations must propagate and replicate correctly on the whole stack, this design not only increases complexity, but it assumes that the wallet networks state is exactly what the dApp is expecting.

 - Delegate defaultChainId to WalletRequestHandler creator
 - Remove defaultChainId from WalletRequestHandler
 - Handle `setDefaultNetwork` using prompter
 - Remove assumptions of networks order from `Wallet`
 - Remove side-effects of `isDefaultChain` on network sorting
 - Unite `testnetNetworks` and `mainnetNetworks`, use a single set of networks

Notice: dapp2 integration tests fail on the cli but pass on the browser, it seems either a previous test messing with the state of that test or a discrepancy between the node runtime and the browser, the tests have been disabled

### DIFF
--- a/packages/provider/src/wallet.ts
+++ b/packages/provider/src/wallet.ts
@@ -371,10 +371,7 @@ export class Wallet implements WalletProvider {
       throw new Error('networks have not been set by session. connect first.')
     }
     // default chain id is first one in the list, by design
-    const network = this.networks[0]
-    if (!network.isDefaultChain) {
-      throw new Error('expecting first network in list to be default chain')
-    }
+    const network = this.networks.find((n) => n.isDefaultChain) ?? this.networks[0]
     return network.chainId
   }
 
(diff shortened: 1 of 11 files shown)


packages/provider/src/wallet.ts
```diff
@@ -619,7 +617,9 @@ export class Wallet implements WalletProvider {
 
     // an extra override for convenience
     if (this.config.networkRpcUrl) {
-      this.networks[0].rpcUrl = this.config.networkRpcUrl
+      // TODO: Is this okay? why do we need to overwride the rpc here?
```
pkieltyka, line 620: we allow the dapp to override it they can have preference over the node-provider, as well if they are a very high usage dapp. In general, the user could change the rpcUrl on the wallet side itself, however, as a Dapp developer who uses 0xsequence directly, to make it easy for them and their users, we allow the dapp to override the rpcUrl here as well

packages/provider/src/wallet.ts
```diff
@@ -619,7 +617,9 @@ export class Wallet implements WalletProvider {
...
+      const defualtChain = this.networks.find((n) => n.isDefaultChain)
+      if (defualtChain) defualtChain.rpcUrl = this.config.networkRpcUrl
```
PhABC, line 622: Typo ; should be `defaultChain` not `defualtChain`

Agusx1211:
Closed in favor of using the existing interfaces on the Networks refactor on wallet-webapp

PR #189, opened 2021-09-08
multicall - Fix reporting of error in MulticallProvider

NicolasMahe:
This PR fixes the reporting of errors in MulticallProvider by returning errors in the callback's error parameter.

I left the error in the callback `JsonRpcResponse` payload to prevent possible side effects. **Please confirm** if the error needs to be removed from this `JsonRpcResponse` payload.

I added checks to the error returned in the unit-test to assert the correct error is returned (it was returning `Cannot read property 'toHexString' of undefined` instead of the actual error).

Here are the logs of the tests before the fix:
```
  1) Multicall integration
       Ether.js provider wrapper
         Handle errors
           Should retry after failing to execute using batch:

      AssertionError: expected promise to be rejected with an error including 'VM Exception while processing transaction: revert CallReceiverMock#testCall: REVERT_FLAG' but got 'Cannot read property \'toHexString\' of undefined'
      + expected - actual

      -Cannot read property 'toHexString' of undefined
      +VM Exception while processing transaction: revert CallReceiverMock#testCall: REVERT_FLAG



  2) Multicall integration
       Ether.js provider wrapper (without proxy)
         Handle errors
           Should retry after failing to execute using batch:

      AssertionError: expected promise to be rejected with an error including 'VM Exception while processing transaction: revert CallReceiverMock#testCall: REVERT_FLAG' but got 'Cannot read property \'toHexString\' of undefined'
      + expected - actual

      -Cannot read property 'toHexString' of undefined
      +VM Exception while processing transaction: revert CallReceiverMock#testCall: REVERT_FLAG
```

### DIFF
--- a/packages/multicall/src/providers/provider.ts
+++ b/packages/multicall/src/providers/provider.ts
@@ -74,7 +74,7 @@ export class MulticallProvider extends ethers.providers.BaseProvider {
   }
 
   private callback(req: JsonRpcRequest, callback: JsonRpcResponseCallback, resp: any, err?: any) {
-    callback(undefined, {
+    callback(err, {
       jsonrpc: JsonRpcVersion,
       id: req.id!,
       result: resp,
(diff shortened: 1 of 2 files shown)


Agusx1211:
Good catch, thanks!

PR #207, opened 2021-11-18
relayer: add Relayer.simulate

attente:
Depends on https://github.com/0xsequence/stack/pull/133.


packages/relayer/src/index.ts
```diff
@@ -5,6 +5,8 @@ import { WalletConfig } from '@0xsequence/config'
 import { proto } from './rpc-relayer'
 
 export interface Relayer {
+  // simulate returns the execution results for a list of transactions.
+  simulate(wallet: string, ...transactions: Transaction[]): Promise<TransactionResult[]>
```
pkieltyka, line 9: should we rename `TransactionResult` to `TxnSimulateResult` or `SimulateResult` ..?

PR #209, opened 2021-11-23
Expose injected transactions

Agusx1211:
- Move building of stub signature to client instead of relayer
- - This allows the client to still generate the stub signature even when the relayed transactions it's nested inside a GuestModule call
- Includes the wallet deployment, if any, at gas estimation
- Split `buildFinalTransaction` on Account.ts
- - Allows to simulate gas estimation for a transaction that includes a lazy configuration update


packages/relayer/src/base-relayer.ts
```diff
@@ -58,62 +58,95 @@ export class BaseRelayer {
...
+    signature: string | Promise<string> | DecodedSignature | Promise<DecodedSignature> | undefined,
     ...transactions: Transaction[]
-  ): Promise<{ to: string, data: string  }> { //, gasLimit?: ethers.BigNumberish }> {
+  ): Promise<{ to: string, transactions: Transaction[], top: { nonce: ethers.BigNumberish, signature: ethers.BytesLike } }> {
     const walletAddress = addressOf(config, context)
-    const walletInterface = new Interface(walletContracts.mainModule.abi)
 
-    const encodedSignature = (async () => {
+    // Solve wallet signature
+    const encodedSignature = await (async () => {
```
attente, line 70: Is there a reason to `await` this `async` IIFE?

PR #215, opened 2021-12-13
wallet: add CallbackRemoteSigner

attente:
Co-authored-by: Corban Riley <corbanbrook@gmail.com>

### DIFF
--- a/packages/wallet/src/remote-signers/callback-remote-signer.ts
+++ b/packages/wallet/src/remote-signers/callback-remote-signer.ts
@@ -1,0 +1,19 @@
+import { BytesLike } from 'ethers'
+import { ChainIdLike } from '@0xsequence/network'
+import { RemoteSigner } from './remote-signer'
+
+export type SignCallback = (message: BytesLike, data?: BytesLike, chainId?: ChainIdLike) => Promise<string>
+
+export class CallbackRemoteSigner extends RemoteSigner {
+  constructor(public readonly address: string, public readonly callback: SignCallback) {
+    super()
+  }
+
+  getAddress(): Promise<string> {
+    return Promise.resolve(this.address)
+  }
+
+  signMessageWithData(message: BytesLike, data?: BytesLike, chainId?: ChainIdLike): Promise<string> {
+    return this.callback(message, data, chainId)
+  }
+}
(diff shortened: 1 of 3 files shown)


packages/wallet/src/remote-signers/callback-remote-signer.ts
```diff
@@ -0,0 +1,19 @@
+import { BytesLike } from 'ethers'
+import { ChainIdLike } from '@0xsequence/network'
+import { RemoteSigner } from './remote-signer'
+
+export type SignCallback = (message: BytesLike, data?: BytesLike, chainId?: ChainIdLike) => Promise<string>
```
pkieltyka, line 5: I think we should change the design a bit. We need both ability to sign a message and also sign a transaction. I propose using an interface instead of a single type, how about:

```typescript
export interface CallbackSigner {
  signMessage(message: MessageToSign): Promise<string>
  signTransaction(txn: TransactionRequest, chaindId?: number): Promise<string>
}
```

I'm also not sure "CallbackSigner" is the right name either. It's just a "Signer". If you notice, we have two places which are similar to this concept:
1. https://github.com/0xsequence/sequence.js/blob/master/packages/wallet/src/signer.ts#L18
2. https://github.com/0xsequence/sequence.js/blob/master/packages/provider/src/transports/wallet-request-handler.ts#L705-L706

I don't think we need yet another interface, instead, we should prob move the sign methods out of the wallet/signer.ts into its own type within that file, .. ie. `interface Signing` perhaps, and then `Signer` can extend that interface.

packages/wallet/src/remote-signers/callback-remote-signer.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export class CallbackRemoteSigner extends RemoteSigner {
```
pkieltyka, line 7: similar to above comment, we should think about the right name for this.. and make sure it `implements Signing` or whatever we name the interface above

pkieltyka (changes requested):

packages/wallet/src/remote-signers/callback-remote-signer.ts
attente: You are probably right. Thinking about it, this PR might not even be needed because what you describe is already covered by the `LocalRemoteSigner` here: https://github.com/0xsequence/sequence.js/blob/597282024a4be39057d2a22f7af7757ab18f48ff/packages/wallet/src/remote-signers/local-remote-signer.ts#L9

And the interface you're describing is basically covered by ethers.js' abstract `Signer` type here: https://github.com/ethers-io/ethers.js/blob/eb432aa1f44ad2cc268d000b266eae9b03db1d17/packages/abstract-signer/src.ts/index.ts#L61-L81

I actually don't think any changes are necessary on sequence.js side any more. We should just implement this interface directly in wallet-webapp without worrying about callbacks.

PR #218, opened 2021-12-20
Confirm wallet deployment prompt

tolgahan-arikan:
- Add checks to signing cases for wallet deployment and config statuses.
- Add prompt for wallet deployment.


pkieltyka:
great work, LGTM :D

there is one failing test though, is this just a CI blip, or missing some other work?

packages/provider/src/transports/wallet-request-handler.ts
```diff
@@ -252,43 +249,72 @@ export class WalletRequestHandler implements ExternalProvider, JsonRpcHandler, P
...
+              throw new Error(`WalletRequestHandler: wallet is not deployed for chainId ${chainId}`)
+            }
+            if (!isUpToDate) {
+              throw new Error(`WalletRequestHandler: wallet config is not up to date for chainId ${chainId}`)
+            }
             sig = await signer.signMessage(ethers.utils.arrayify(message), chainId)
           } else {
-            // prompt user to provide the response
-            sig = await this.prompter.promptSignMessage({ chainId: chainId, message: message })
+            // check if wallet is deployed and up to date, if not, prompt user to deploy
```
pkieltyka, line 299: btw, for this block of code, once `promptConfirmWalletDeploy` is executed, will it then ask the user to the sign the message automatically? or will the dapp have to prompt to sign the message again..?

ideally, once it executes the deployment, it will then ask to user the sign the message which initiated this execution

tolgahan-arikan: yep it does, a few lines below

```
if (promptResultForDeployment) {
    sig = await this.prompter.promptSignMessage({ chainId: chainId, message: message })
}
```

packages/provider/src/transports/wallet-request-handler.ts
```diff
@@ -252,43 +249,66 @@ export class WalletRequestHandler implements ExternalProvider, JsonRpcHandler, P
...
           // TODO:
           // if (process.env.TEST_MODE === 'true' && this.prompter === null) {
           if (this.prompter === null) {
             // prompter is null, so we'll sign from here
             sig = await signer.signMessage(ethers.utils.arrayify(message), chainId)
           } else {
-            // prompt user to provide the response
-            sig = await this.prompter.promptSignMessage({ chainId: chainId, message: message })
+            // check if wallet is deployed and up to date, if not, prompt user to deploy
+            let isDeployed: boolean
```
Agusx1211, line 277: Can we move this block of code to a standalone function so we don't have it duplicated on `eth_sign` and `eth_signTypedData`? this method could just return `isUpToDate` and assume that the wallet not being deployed is just another way of not being up to date

packages/provider/src/transports/wallet-request-handler.ts
```diff
@@ -320,12 +340,47 @@ export class WalletRequestHandler implements ExternalProvider, JsonRpcHandler, P
...
+              }
+            } else {
+              // if no chainId is provided, we'll assume the wallet is auth chain wallet
+              isDeployed = true
+              isUpToDate = true
+            }
+
+            let promptResultForDeployment = true
+            if (!isDeployed || !isUpToDate) {
+              const promptResult = await this.prompter.promptConfirmWalletDeploy(chainId)
```
Agusx1211, line 368: Maybe this can also be moved to a shared function

Agusx1211 (changes requested):

Agusx1211 (approved):
LGTM!

PR #223, opened 2022-01-14
config: batch eth_getCode calls when building stub signatures

attente:


pkieltyka:
wow, love how simple this is.. :P 

a few thoughts..

1. I think multicall has two different approaches, it will auto batch calls, but also can be explicit about it, I think explicit might be better / more performant..? but maybe doesn't make a difference
2. We are still using eth_getCode, this is fine I guess, but it could be a lot of data in the call..?

attente:
So my (limited) understanding on reviewing the multicall interface is that it automatically batches all calls that happen within a small time window which defaults to 50 ms: https://github.com/0xsequence/sequence.js/blob/2cbeed0ee11ce3bc60667dbbd9bb70125120fef3/packages/multicall/src/multicall.ts#L33. There isn't (afaict) a way to surface the idea of "I want to batch a known set of requests all at once, and then explicitly flush those requests when I know I'm done". @PhABC, @Agusx1211 do you know if there's a way to do this?

For the second point, I reconsidered using the code hash for a few reasons:
1. Almost all of the signers on a sequence wallet are going to be EOA signers, so I think eth_getCode should be faster than the codehash since they're just returning empty data. Really only one signer is going to be non-EOA which is the guard signer. At some point this might change, but this leads to the second point:
2. If a non-EOA signer happens to be a Sequence wallet, then the code implementing it is incredibly small, thanks to Agustin and Philippe's work on optimizing it. It's only 26 bytes long, which is even shorter than a keccak256 hash.

PR #225, opened 2022-01-18
config: cache wallet configs

attente:


packages/config/src/config.ts
```diff
@@ -92,14 +92,29 @@ export const addressOf = (salt: WalletConfig | string, context: WalletContext, i
   return addressOf(imageHash(salt), context)
 }
 
+// cached wallet configs indexed by image hash
+// wallet configs must be sorted using sortConfig
+const cachedConfigs: Map<string, WalletConfig> = new Map()
+
```
pkieltyka, line 98: maybe we should add another method called, `setCachedConfig(xxx)` so that we could get/set this from the wallet-webapp from local storage, so it would survive reloads?

in other parts in findConfig, if we load something from cache, we should check still the on-chain "imageHash" compared to our latest cached value, this way we'll know if our cache is expired..

what do you think?

attente: It's possible to move this cache from memory to local storage, but in that case, we will have to consider adding an eviction strategy so that the cache isn't unbounded.

> in other parts in findConfig, if we load something from cache, we should check still the on-chain "imageHash" compared to our latest cached value, this way we'll know if our cache is expired..

I think I might be misunderstanding this remark. `findConfigForImageHash` has a specific purpose, to find the config for a specific image hash. If you're referring to `findCurrentConfig`, then it already makes a request for the on-chain image hash.

packages/config/src/cache.ts
```diff
@@ -0,0 +1,48 @@
...
+const configKey = (imageHash: string) => `@sequence.config.${imageHash}`
+
+let storage: {
+  getItem(key: string): string | null
+  setItem(key: string, value: string): void
+  removeItem(key: string): void
+}
+
+try {
+  storage = localStorage
```
pkieltyka, line 15: nice, looks good to try localStorage and have fallback.

does this work though..? I wonder if you need `(window as any).localStorage` ..? if this works though, looks great

attente: Yeah, so this works, but I'm not sure how idiomatic it is. But config tests without local storage still work, and it runs in browser too. And no `any` cast is necessary here.

attente:
### Current behaviour prior to this PR

The wallet calls `findConfigForImageHash` repeatedly for the same image hashes, over and over and over again. It's noticeable when the user lands on the wallet index route, but also noticeable when you visit the wallet deployment route on a current master build.



Each one of these produces an `eth_getLogs` call to the node.

### New behaviour as proposed by this PR

This PR adds an LRU local storage cache so that any time we request the wallet config for a specific image hash multiple times, it only does it the first time, and caches the config.

With 10 configs in the cache, it looks like this:



After removing a session from my wallet, which generates a new wallet config, this happens (image hash 0xb341ce9b... is the LRU here, so it evicts that wallet config from the cache):



Note that the console shows only two cache misses total that result in `eth_getLogs` requests to the node.

PR #226, opened 2022-02-02
relayer: fee quote support

attente:


packages/relayer/src/index.ts
```diff
@@ -44,7 +53,20 @@ export { proto as RpcRelayerProto } from './rpc-relayer'
 export type SimulateResult = proto.SimulateResult
 export type FeeOption = proto.FeeOption
 
+export interface FeeQuote {
```
pkieltyka, line 56: nice work on this PR! looks slick.

how come for this type you use keys of _tag and _quote..?

attente: My rationale was that IMO the fee quote itself is simply an opaque object that a sequence.js consumer doesn't ever need to inspect because all it is ever used for is to forward it back to the same relayer when the transaction is ready to dispatch. So the type of the fee quote could effectively be the TypeScript `unknown`. However using `quote: unknown` doesn't have strong enough type safety because then any arbitrary JS value can be passed in. So I took this approach to create a nominal opaque type that also maintains type safety.

attente: Thanks for asking about this. I added a comment to clarify this in the code: https://github.com/0xsequence/sequence.js/pull/226/commits/f06cd567dd9c42773ae7aed0a064f5272bd4604a.

PR #231, opened 2022-02-07
Setting target to 0x0 when empty to during SequenceTxAbiEncode

corbanbrook:


attente (approved):
LGTM

`sequenceTxAbiEncode` is always used specifically for encoding transactions for `execute` and `selfExecute` calls. So it should be safe to assume that the `to` address would have been set properly by that point if it were an ordinary non-deploying transaction. Thanks for debugging this!

PR #244, opened 2022-03-18
:arrow_up: attempting to force node-fetch to be 2.6.7 or higher

maschad:
Closes https://github.com/horizon-games/issue-tracker/issues/7167

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,9 @@
     "wait-on": "^6.0.0",
     "wsrun": "^5.2.4"
   },
-  "resolutions": {},
+  "resolutions": {
+    "node-fetch": "^2.6.7"
+  },
   "workspaces": [
     "packages/*"
   ],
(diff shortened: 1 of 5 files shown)


attente:
I wonder if switching to isomorphic-fetch is going to be an issue for you in React Native:



The issue in question: https://github.com/matthew-andrews/isomorphic-fetch/issues/125.

pkieltyka:
Good news if that node 17 implements native support for fetch() so it’s a matter of time before we can do without any of these polyfill libs 

Sounds like we should stick to cross-fetch and upgrade to the latest major, as well node-fetch if it’s a sub-deb. Using the latest majors of both should work just fine

maschad:
> Good news if that node 17 implements native support for fetch() so it’s a matter of time before we can do without any of these polyfill libs
> 
> Sounds like we should stick to cross-fetch and upgrade to the latest major, as well node-fetch if it’s a sub-deb. Using the latest majors of both should work just fine

I would agree with that exception that should a new vulnerability arise in `node-fetch` then we would be beholden to their maintainer until they upgrade the dependency or facilitate minor/major upgrades. This could leave us exposed.

PR #255, opened 2022-04-18
Add prefix for EIP-191 signature before hashing (Alternative)

Agusx1211:
Replaces #249 

Implements EIP-191 prefixing of message, it allow Sequence wallet to login on Opensea among other dapps.

In contrast with the #249 implementation this PR implements the prefixing at the wallet-request-handler level, this ensures low-level signing of non EIP191 messages is still available and doesn't introduce complexity at the digest utility methods.


PhABC (approved):
LGTM

Need some testing on it. Can we merge + release it? Then I'll ask Nikki to do a round of testing on a few signing related things on a dev environment.

PR #258, opened 2022-04-26
Generalized dapp connections

tolgahan-arikan:
- Add ConnectedDapp and DappConnectionType
- Use for all types of connections instead of only in extension mode
- Add origin to accountsChanged to be able to disconnect a specific dapp


packages/provider/src/transports/base-provider-transport.ts
```diff
@@ -222,14 +226,17 @@ export abstract class BaseProviderTransport implements ProviderTransport {
...
 
-    // ACCOUNTS_CHANGED -- when a user logs in or out
+    // ACCOUNTS_CHANGED -- when a user logs in, logs out or disconnects an app
     if (message.type === EventType.ACCOUNTS_CHANGED) {
-      this.accountsChangedPayload = undefined
-      if (message.data && message.data.length > 0) {
-        this.accountsChangedPayload = ethers.utils.getAddress(message.data[0])
-        this.events.emit('accountsChanged', [this.accountsChangedPayload])
+      this.accountsChangedPayload = { accounts: [] }
+      if (message.data && (message.data.length > 0 || message.data.accounts?.length > 0)) {
```
attente, line 232: Is it possible for `message.data.length === 0` when `message.data.accounts?.length > 0`?

packages/provider/src/transports/base-provider-transport.ts
```diff
@@ -222,14 +226,17 @@ export abstract class BaseProviderTransport implements ProviderTransport {
...
+        this.accountsChangedPayload = {
+          accounts: [ethers.utils.getAddress(message.data[0] || message.data.accounts[0])],
```
attente, line 234: It would be nice if we had better typings that discriminate based on the `message.type`, but I guess that makes more sense in a separate PR.

packages/provider/src/transports/base-wallet-transport.ts
```diff
@@ -420,26 +417,37 @@ export abstract class BaseWalletTransport implements WalletTransport {
...
+            session: await this.walletRequestHandler.walletSession()
+          })
+        } else {
+          this.notifyOpen({
+            sessionId: this._sessionId,
+            error: ERROR_MESSAGES.DAPP_NOT_CONNECTED
+          })
+          logger.error(`dapp origin ${this.appOrigin} not found in connected dapps`)
+          return false
+        }
```
attente, line 438: I'm maybe not the best person to review this, but is there some explanation why it's these three conditions that call `this.notifyOpen`? Also, is the error message here correct for the conditions needed to return an error?

packages/provider/src/transports/wallet-request-handler.ts
```diff
@@ -796,6 +827,18 @@ export class WalletRequestHandler implements ExternalProvider, JsonRpcHandler, P
     }
     return false
   }
+
+  private saveConnectedDapp(connectedDapp: ConnectedDapp) {
+    const connectedDappsKey = LOCAL_STORAGE_KEYS.CONNECTED_DAPPS
+    const data = window.localStorage.getItem(connectedDappsKey)
+    const connectedDapps: ConnectedDapp[] = data ? JSON.parse(data) : []
+    if (!connectedDapps.map(dapp => dapp.origin).includes(connectedDapp.origin)) {
```
attente, line 835: I think I'd reverse the order of the if-else blocks here and do `connectedDapps.some(dapp => dapp.origin === connectedDapp.origin)`, but feel free to ignore since it's really minor.

packages/provider/src/utils.ts
```diff
@@ -162,3 +160,11 @@ export class LocalStore<T extends Object = string> {
     window.localStorage.removeItem(this.key)
   }
 }
+
+export const isDappConnected = (origin: string | undefined): boolean => {
+  if (!origin) return false
+  const connectedDappsKey = LOCAL_STORAGE_KEYS.CONNECTED_DAPPS
+  const data = window.localStorage.getItem(connectedDappsKey)
+  const connectedDapps: ConnectedDapp[] = data ? JSON.parse(data) : []
+  return connectedDapps.map(dapp => dapp.origin).includes(origin)
```
attente, line 169: Same thing with `saveConnectedDapp`, but again feel free to ignore.

attente (approved):
It all looks pretty sane to me, but might not hurt to get another pair of eyes on this from someone more experienced.

packages/provider/src/transports/base-wallet-transport.ts
tolgahan-arikan: Yeah I understand it's not clear as it is, need to add more comments here.

So previously we weren't checking if a dapp is connected at all. This change checks if a dapp connected with `isDappConnected`, which is the first and the main condition. Second one is to allow if this request came from extension, and last if there is no wallet request handler, in which case a dapp would not be added to connected dapps at all so we can allow connecting directly.

And about the error, in extension 'not being connected' is handled earlier, so main case this error is needed is for dapps that integrate with sequence.js directly. If user disconnects a dapp in settings, this error allows us to clear the session from that app next time they try to open the wallet.

packages/provider/src/transports/base-provider-transport.ts
tolgahan-arikan: Just did a refactor here and removed that part as it wasn't necessary anymore. Moved origin to ProviderMessage so that it won't pollute the data itself.

packages/provider/src/transports/base-provider-transport.ts
```diff
@@ -222,14 +226,17 @@ export abstract class BaseProviderTransport implements ProviderTransport {
...
-    // ACCOUNTS_CHANGED -- when a user logs in or out
+    // ACCOUNTS_CHANGED -- when a user logs in, logs out or disconnects an app
     if (message.type === EventType.ACCOUNTS_CHANGED) {
-      this.accountsChangedPayload = undefined
+      this.accountsChangedPayload = { accounts: [] }
       if (message.data && message.data.length > 0) {
-        this.accountsChangedPayload = ethers.utils.getAddress(message.data[0])
-        this.events.emit('accountsChanged', [this.accountsChangedPayload])
+        this.accountsChangedPayload = {
+          accounts: [ethers.utils.getAddress(message.data[0])],
```
attente, line 234: Is there a reason it's only the first account getting added here?

packages/provider/src/transports/wallet-request-handler.ts
```diff
@@ -709,16 +742,15 @@ export class WalletRequestHandler implements ExternalProvider, JsonRpcHandler, P
...
     this.events.emit('connect', connectDetails)
     if (connectDetails.session?.accountAddress) {
-      this.events.emit('accountsChanged', [connectDetails.session?.accountAddress])
+      this.events.emit('accountsChanged', [connectDetails.session?.accountAddress], origin)
     }
   }
 
-  notifyDisconnect() {
-    this.events.emit('accountsChanged', [])
-    this.events.emit('networks', [])
```
attente, line 721: This removal is intentional?

PR #263, opened 2022-05-03
Lazy octopus 2

Agusx1211:


packages/0xsequence/tests/browser/wallet-provider/dapp2.test.ts
```diff
@@ -64,23 +60,8 @@ export const tests = async () => {
...
-    // The non-default chain's chain ID is 31337
-    assert.equal(networks.filter(network => !network.isDefaultChain)[0].chainId, 31337, 'non-default chain id is 31337')
-
-    // The auth chain's chain ID is 31338
-    assert.equal(networks.filter(network => network.isAuthChain)[0].chainId, 31338, 'auth chain id is 31338')
-
-    // The non-auth chain's chain ID is 31337
-    assert.equal(networks.filter(network => !network.isAuthChain)[0].chainId, 31337, 'non-auth chain id is 31337')
+    assert.true(networks[0].isDefaultChain, 'network0 is defaultChain')
+    assert.true(!networks[1].isDefaultChain, 'network1 is not defaultChain')
```
attente, line 64: Is there a reason the test now assumes the first is the default? Isn't it more appropriate to just check that there is exactly one default chain?

packages/auth/src/session.ts
```diff
@@ -257,48 +261,22 @@ export class Session {
...
-          if (!val.config) throw Error("Can't find latest config")
-          return authWallet.wallet
-            .useConfig(val.config!)
-            .sign(proof.messageDigest())
-            .then(signature => {
-              const decodedSignature = decodeSignature(signature)
-              const totalWeight = decodedSignature.signers.filter(isDecodedSigner).reduce((totalWeight, signer) => totalWeight + signer.weight, 0)
-              if (totalWeight < decodedSignature.threshold) {
-                throw Error(`insufficient signing power, need ${decodedSignature.threshold}, have ${totalWeight}`)
-              }
```
attente, line 290: Is the signature guaranteed to meet threshold? This was a workaround for https://github.com/0xsequence/wallet-webapp/pull/328 IIRC, so it's nice if it could be removed, but just want to confirm that lazy octopus fixes this.

PR #278, opened 2022-06-07
Add origin info for accountsChanged event to handle it per dapp

tolgahan-arikan:


packages/provider/src/transports/base-provider-transport.ts
```diff
@@ -37,7 +37,7 @@ export abstract class BaseProviderTransport implements ProviderTransport {
 
   protected openPayload: { sessionId?: string; session?: WalletSession } | undefined
   protected connectPayload: ConnectDetails | undefined
-  protected accountsChangedPayload: string | undefined
+  protected accountsChangedPayload: { accounts: string[]; origin?: string } | undefined
```
pkieltyka, line 40: good catch, I also checked EIP1193, and this looks like the correct format to return (for `{accounts: []string}` and since origin is optional + additive, it wont break anything)

packages/provider/src/transports/base-provider-transport.ts
```diff
@@ -224,12 +224,15 @@ export abstract class BaseProviderTransport implements ProviderTransport {
...
-      this.accountsChangedPayload = undefined
+      this.accountsChangedPayload = { accounts: [] }
       if (message.data && message.data.length > 0) {
-        this.accountsChangedPayload = ethers.utils.getAddress(message.data[0])
-        this.events.emit('accountsChanged', [this.accountsChangedPayload])
+        this.accountsChangedPayload = {
+          accounts: [ethers.utils.getAddress(message.data[0])],
+          origin: message.origin
+        }
+        this.events.emit('accountsChanged', this.accountsChangedPayload.accounts, this.accountsChangedPayload.origin)
```
pkieltyka, line 233: I wonder if we should just emit `this.accountsChangedPayload` ? I suppose this could be a breaking change which is the problem, is it more correct according to EIP1193 and what other consumers are expecting?

packages/provider/src/transports/base-wallet-transport.ts
```diff
@@ -206,11 +206,12 @@ export abstract class BaseWalletTransport implements WalletTransport {
     })
   }
 
-  notifyAccountsChanged(accounts: string[]) {
+  notifyAccountsChanged(accounts: string[], origin?: string) {
     this.sendMessage({
       idx: -1,
       type: EventType.ACCOUNTS_CHANGED,
-      data: accounts
+      data: accounts,
```
pkieltyka, line 213: we may want to return `data: { accounts, origin }` .. thoughts?

pkieltyka (commented):
nice work, sent some questions

packages/provider/src/transports/base-provider-transport.ts
tolgahan-arikan: yeah this is mainly about `TypedEventEmitter.emit` method and needing to append things so that i will not break current usages. The consumers are dapps connected and the wallet webapp.

packages/provider/src/transports/base-wallet-transport.ts
tolgahan-arikan: yeah I tried this one, this one breaks communications with dapps since they accept an array here.

pkieltyka: Gotcha. Could you open issue for us in our issuer-tracker that we should review this event in the future? When we do release of lazy octopus we’ll want all dapps to upgrade, so we can make the change then for what potentially may be more correct

packages/provider/src/wallet.ts
```diff
@@ -244,7 +253,7 @@ export class Wallet implements WalletProvider {
 
     if (
       this.isConnected() &&
-      this.isSiteConnected(options?.origin) &&
+      (await this.isSiteConnected(options?.origin)) &&
```
attente, line 256: nice catch

PR #281, opened 2022-07-13
singleton helpers

pkieltyka:


packages/provider/src/wallet.ts
```diff
@@ -717,6 +717,10 @@ export const DefaultProviderConfig: ProviderConfig = {
 
 let walletInstance: Wallet | undefined
 
+export const configWallet = (network?: string | number, config?: Partial<ProviderConfig>) => {
+  walletInstance = new Wallet(network, config)
```
arilotter, line 721: This isn't safe - if you call this more than once, you'll break sequence, since it'll set up the transports more than once ever.

packages/provider/src/wallet.ts
```diff
@@ -728,6 +732,20 @@ export const useWallet = () => {
   return walletInstance
```
arilotter, line 732: what's the point of this method existing?
it's the same as getWallet, but not safe, because it will sometimes return `undefined`

packages/provider/src/wallet.ts
```diff
@@ -728,6 +732,20 @@ export const useWallet = () => {
...
 }
 
-export const resetWallet = () => {
-  walletInstance = undefined
+export const connectWallet = async (network?: string | number, options?: ConnectOptions & { walletConfig?: Partial<ProviderConfig> }) => {
+  if (!walletInstance) {
+    walletInstance = getWallet(network, options?.walletConfig)
+  }
+  if (network && options) {
+    options.networkId = network
```
arilotter, line 740: passing network here isn't the same as setting the default network ID in configWallet or getWallet - just as a note.
also don't see an advantage for having this over just making people call both get and connect methods.

packages/provider/src/wallet.ts
```diff
@@ -728,6 +732,20 @@ export const useWallet = () => {
...
+  }
+  const connectDetails = await walletInstance.connect(options)
+  return { wallet: walletInstance, connectDetails }
+}
+
+export const disconnectWallet = () => {
```
arilotter, line 746: this is never safe, we can't dispose of the wallet, otherwise getWallet will crash when we call it next, because we can't duplicate the transport

arilotter (changes requested):

packages/provider/src/wallet.ts
pkieltyka: this is the same as we have right now which is `const wallet = new sequence.Wallet(network, config)` .. and people survive. Dev's should know you don't config all the time, just at init time. In fact, in many cases it won't be called, but for advanced usage with singleton, we need a way to recreate the singleton when necessary

packages/provider/src/wallet.ts
pkieltyka: Agree, `useWallet` just confuses it by having multiple use for similar/same thing

packages/provider/src/wallet.ts
pkieltyka: I think its necessary for expected use, and shouldn't break things.. cuz people may want to call connectWallet() multiple times will using the same instance

packages/provider/src/wallet.ts
pkieltyka: this is the same as `wallet.disconnect()` which a dapp certainly may want to use. it's just a short-hand.. but we could remove it and force devs to just .. `sequence.getWallet().disconnect()` .. I think that is fine too, so yea will remove this method

packages/provider/src/wallet.ts
arilotter: right, we're trying to work around this problem with this singleton stuff
it's never safe to create more than one wallet instance
there's no "recreating the singleton" that works with the current code setup, this method does not function as intended.

PR #285, opened 2022-07-25
add unreal transport

arilotter:


packages/provider/src/transports/unreal-transport/overridelogs.ts
```diff
@@ -0,0 +1,36 @@
...
+      t.errorfromjs(`${side}: ${stringify(args)}`)
+    }
+    window.logsOverriddenForUnreal = true
+  }
+}
+
+function stringify(things: unknown[]): string {
+  return `${things
+    .map(a => (typeof a === 'object' ? (a instanceof Error ? a.message : JSON.stringify(a)) : String(a)))
+    .join(' ')}`
```
attente, line 35: Why is this a template string?

packages/provider/src/transports/unreal-transport/unreal-message-handler.ts
```diff
@@ -0,0 +1,130 @@
...
+          this.notifyClose({ message: err } as ProviderRpcError)
+          window.close()
+        }
+      })
+      .catch(e => {
+        const err = `failed to open to network ${session?.networkId}, due to: ${e}`
+        logger.error(err)
+        this.notifyClose({ message: err } as ProviderRpcError)
+        window.close()
+      })
```
attente, line 79: Bit nitpicky, but why not use `await` here?

arilotter: because the register function doesn't wait for the .open event, it executes asynchronously. we don't return this promise

packages/provider/src/transports/unreal-transport/unreal-message-provider.ts
```diff
@@ -0,0 +1,121 @@
...
+      console.log('wallet already open!')
+      return
+    }
+
+    console.log('opening wallet!')
+    // Instantiate new walletURL for this call
+    const walletURL = new URL(this.walletURL.href)
+    const windowSessionParams = new WindowSessionParams()
+
+    if (path && path !== '') {
```
attente, line 77: I think `if (path) {` is sufficient here.

packages/provider/src/transports/unreal-transport/unreal-message-provider.ts
```diff
@@ -0,0 +1,121 @@
...
+      walletURL.pathname = path.toLowerCase()
```
attente, line 78: Is it correct to always lower case the path? Shouldn't we use the path as provided by the dapp?

attente (approved):

PR #291, opened 2022-09-19
provider: forward delegateCall and revertOnError transaction fields

attente:


packages/provider/src/provider.ts
```diff
@@ -479,7 +479,10 @@ const hexlifyTransaction = (
   }
   ethers.utils.checkProperties(transaction, allowed)
 
-  const result: { [key: string]: any } = {}
+  const result: { [key: string]: any } = {
+    delegateCall: (transaction as any).delegateCall,
```
pkieltyka, line 483: say `(transaction as any).delegateCall` is undefined.. on result we will be creating keys with undefined values...

should we instead do like..

```ts
if ((transaction as any).delegateCall) {
  result.delegateCall = (transaction as any).delegateCall
}
// etc..
```

thoughts?

packages/provider/src/provider.ts
```diff
@@ -498,6 +498,12 @@ const hexlifyTransaction = (
     }
     result[key] = ethers.utils.hexlify((<any>transaction)[key])
   })
+  ;['delegateCall', 'revertOnError'].forEach(key => {
+    const value = (transaction as any)[key]
+    if (value !== undefined) {
```
pkieltyka, line 503: maybe just if (value) .. it will check undefined and null..

PR #296, opened 2022-10-04
relayer: remove deprecated api and other improvements

attente:


attente:
This branch should be rebased, not squashed.

pkieltyka (approved):

PR #307, opened 2022-10-19
add eip 191 exceptions

arilotter:


packages/provider/src/eip191exceptions.ts
```diff
@@ -0,0 +1,123 @@
...
+  // See https://gov.0x.org/t/zeip-proposal-fix-v3-eip-191-non-compliance-when-validating-eip-1271-signatures/3396 for more info.
+  { name: '0x v3 Exception', predicate: isZeroExV3Order }
+]
+
+// try to interpret bytes as abi-encoded 0x v3 OrderWithHash -
+// see https://github.com/0xProject/0x-protocol-specification/blob/master/v3/v3-specification.md
+export function isZeroExV3Order(bytes: Uint8Array): boolean {
+  const abi = new ethers.utils.Interface(ZeroXV3EIP1271OrderWithHashAbi)
+  try {
+    abi.decodeFunctionData('OrderWithHash', bytes)
```
Agusx1211, line 24: Is this the best we can do to filter out 0x orders? maybe there is also something inside the parsed order that we can also validate

packages/provider/src/eip191exceptions.ts
```diff
@@ -0,0 +1,123 @@
...
+export function messageIsExemptFromEIP191Prefix(message: Uint8Array, origin: string | undefined): boolean {
+  return EIP_191_PREFIX_EXCEPTIONS.some(e => e.predicate(message, origin))
+}
+
+const EIP_191_PREFIX_EXCEPTIONS: Array<{
+  name: string
+  predicate: (message: Uint8Array, origin?: string) => boolean
+}> = [
+  // NOTE: Decentraland does not support 191 correctly.
+  { name: 'Decentraland Exception', predicate: (_message, origin) => origin === 'https://play.decentraland.org' },
```
Agusx1211, line 12: Maybe if we have the advanced predicate we could also validate if the message from DCL is "safe", we could do this by validating that the message DCL is asking us to sign is actually a text message and not an encoded Sequence transaction

Agusx1211 (changes requested):

packages/provider/src/utils.ts
```diff
@@ -15,9 +16,13 @@ export const messageToBytes = (message: BytesLike): Uint8Array => {
   return ethers.utils.toUtf8Bytes(message)
 }
 
-export const prefixEIP191Message = (message: BytesLike): Uint8Array => {
+export const prefixEIP191Message = (message: BytesLike, origin: string | undefined): Uint8Array => {
```
attente, line 19: Is there a reason you opted for `origin: string | undefined` instead of `origin?: string`? Yours will force an API break while the latter doesn't, but I'm not sure if you intended for that to force consumers to consider this.

packages/provider/src/eip191exceptions.ts
arilotter: sure, can easily add a requirement that the message decodes as UTF-8 - could even make it a string matcher rather than an origin matcher - does DCL require us to do messages other than the login sign 191-exempted?

packages/provider/src/eip191exceptions.ts
arilotter: this does decode all the way to a 0x order - could verify that the hash matches the order, though, sound good?

Agusx1211: Thinking harder about this I don't think it's *that* necessary, but it shouldn't hurt. If it isn't much work to add it then it sounds good.

packages/provider/src/eip191exceptions.ts
Agusx1211: I don't think they do, and having those additional checks would limit a lot the kind of damage DCL can do.

packages/provider/src/utils.ts
arilotter: honestly, i would love to remove origin. let me see if i can match DCL messages in another way.

packages/provider/src/eip191exceptions.ts
arilotter: we can't verify the hash is correct, since we don't know what chain they're signing the typed data for, so we can't recreate the same hash

PR #310, opened 2022-11-08
Fix sequence utils address in multicall/README.md

rellfy:


pkieltyka:
thanks @rellfy for the PR

PR #331, opened 2022-12-07
Fix init wallet network selection

tolgahan-arikan:
issue: initWallet creates a Wallet instance and then loads the session, loadSession calls useSession, which sets the network from the persisted session (if there is) and therefore the network passed to initWallet is ignored.

fixes https://github.com/horizon-games/issue-tracker/issues/9131

to test the current broken state, you can check following on demo-dapp

1. initWallet with any network, connect
2. change network in initWallet to another network in demo-dapp, refresh the page, check chainID to see network doesn't actually change


packages/provider/src/wallet.ts
```diff
@@ -248,15 +249,37 @@ export class Wallet implements WalletProvider {
...
       return undefined
     }
+
     try {
       const session = JSON.parse(data) as WalletSession
       if (session) {
-        this.useSession(session, false)
+        // Setting preferredNetwork as default network if it's in session.networks
+        if (preferredNetwork !== undefined) {
+          const preferredNetworkIdNum = parseInt(preferredNetwork as any)
```
attente, line 263: Applying `parseInt` to non-strings is discouraged: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt#using_parseint_on_non-strings.

packages/provider/src/wallet.ts
```diff
@@ -248,15 +249,37 @@ export class Wallet implements WalletProvider {
...
+          const preferredNetworkInConfig = session.networks?.find(
+            n => n.name === preferredNetwork || n.chainId === preferredNetworkIdNum
```
attente, line 265: Can you assign this predicate to a separate variable?

```
const isPreferredNetwork = n => n.name === preferredNetwork || n.chainId === preferredNetworkIdNum
const preferredNetworkInConfig = session.networks?.find(isPreferredNetwork)
```

packages/provider/src/wallet.ts
```diff
@@ -248,15 +249,37 @@ export class Wallet implements WalletProvider {
...
+          const isAlreadyDefaultChain = preferredNetworkInConfig?.isDefaultChain
+
+          if (session.networks && preferredNetworkInConfig && !isAlreadyDefaultChain) {
+            const updatedNetworks = session.networks.map(n => {
+              n.isDefaultChain = false
+              if (n.name === preferredNetwork || n.chainId === preferredNetworkIdNum) {
+                n.isDefaultChain = true
+              }
+              return n
+            })
```
attente, line 276:
```
const updatedNetworks = session.networks.map(n => ({ ...n, isDefaultChain: isPreferredNetwork(n) }))
```

attente (changes requested):
Minor requested changes, but mostly looks good. Thanks for the fix!

attente (approved):

PR #362, opened 2023-04-18
Deploy script

ScreamingHawk:
Added contract verification (tenderly and etherscan) to universal deployer scripts. Added ability to recover funds after deployment. Added static bytecode validation. 

Appreciate any comments / concerns about this approach. 

I'd also like to use solc to generate bytecode from source rather than depend on env vars and typechain. Given we are using compiler settings to verify, this shouldn't be a big ask from integrators.


ScreamingHawk:
Closing this PR. Moving the script to a new repo purely for deployments

PR #368, opened 2023-05-03
backward compatible getWalletState

attente:
Requesting wallet state without a chain ID used to return the state on all chains.


Agusx1211:
I like it, but maybe is worth to add some `v2` methods with the native output of these functions? or maybe we can add them later, not sure really

PR #381, opened 2023-06-28
Remove duplicated defaultChainId on wallet request handler

Agusx1211:
This removes `defaultNetworkId` from the wallet request handler, in its place we use the prompter to query this information directly from the wallet webapp. This avoids duplicating this information, something that could lead to sync issues.

### DIFF
--- a/packages/provider/src/transports/wallet-request-handler.ts
+++ b/packages/provider/src/transports/wallet-request-handler.ts
@@ -30,8 +30,7 @@ const SIGNER_READY_TIMEOUT = 10000
 
 export interface WalletSignInOptions {
   connect?: boolean
-  networks?: NetworkConfig[]
-  defaultNetworkId?: string | number
+  defaultNetworkId?: number
 }
 
 export class WalletRequestHandler implements ExternalProvider, JsonRpcHandler, ProviderMessageRequestHandler {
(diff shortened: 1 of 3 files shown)


packages/provider/src/transports/wallet-request-handler.ts
```diff
@@ -831,13 +826,17 @@ export class WalletRequestHandler implements ExternalProvider, JsonRpcHandler, P
 }
 
 export interface WalletUserPrompter {
+  getDefaultNetwork(): number
```
corbanbrook, line 829: I think we should use the term chainId. ie.`getDefaultChainId` to avoid any disambiguation with NetworkConfigs

packages/provider/src/transports/wallet-request-handler.ts
```diff
@@ -51,19 +50,18 @@ export class WalletRequestHandler implements ExternalProvider, JsonRpcHandler, P
...
-    defaultNetworkId?: string | number
+    networks: NetworkConfig[]
   ) {
     this.account = account
     this.prompter = prompter
     this.networks = networks
+  }
 
-    this.defaultNetworkId = defaultNetworkId ? this.findNetworkID(defaultNetworkId) : networks[0].chainId
+  defaultNetworkId(): number {
```
corbanbrook, line 63: rename to `defaultChainId`

I believe we use the term networkId or the NetworkId type to mean a string | number where the string could be a named identifier from the NetworkConfig:name field, ie. 'mainnet', 'polygon'

we should be clear that this is a chainId

packages/provider/src/transports/wallet-request-handler.ts
```diff
@@ -704,9 +699,9 @@ export class WalletRequestHandler implements ExternalProvider, JsonRpcHandler, P
     this.onConnectOptionsChange?.(options)
   }
 
-  setDefaultNetwork(chainId: string | number): number {
-    this.defaultNetworkId = this.findNetworkID(chainId)
-    return this.defaultNetworkId
+  async setDefaultNetwork(chainId: number): Promise<number> {
```
corbanbrook, line 702: setDefaultChainId

corbanbrook (changes requested):

Agusx1211:
I also fixed `getChainId` on the provider. It previously opened wallet webapp and because of that, it returned the wrong value. Now it just returns whatever is defined as default.

packages/provider/src/transports/wallet-request-handler.ts
```diff
@@ -831,13 +826,17 @@ export class WalletRequestHandler implements ExternalProvider, JsonRpcHandler, P
 }
 
 export interface WalletUserPrompter {
+  getDefaultChainId(): number
```
pkieltyka, line 829: btw, just wondering, why do we need `getDefaultChainId()` on the prompter? The idea of the prompter interface is that these are methods which require "user input" / "user confirmation", so the UI will actually prompt the user to proceed or decline

PR #382, opened 2023-06-28
Util for trimming EIP-191 prefix

tolgahan-arikan:
A utility function to trim EIP-191 signed message prefixes for the purposes of displaying a clean signing request on a frontend.

### DIFF
--- a/packages/provider/src/transports/wallet-request-handler.ts
+++ b/packages/provider/src/transports/wallet-request-handler.ts
@@ -374,7 +374,9 @@ export class WalletRequestHandler implements ExternalProvider, JsonRpcHandler, P
           if (typeof typedDataObject === 'string') {
             try {
               typedData = JSON.parse(typedDataObject)
-            } catch (e) {}
+            } catch (e) {
+              console.warn('error parsing typedData', e)
+            }
           } else {
             typedData = typedDataObject
           }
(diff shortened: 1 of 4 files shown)


packages/provider/src/transports/wallet-request-handler.ts
```diff
@@ -379,7 +379,9 @@ export class WalletRequestHandler implements ExternalProvider, JsonRpcHandler, P
           if (typeof typedDataObject === 'string') {
             try {
               typedData = JSON.parse(typedDataObject)
-            } catch (e) {}
+            } catch (e) {
+              console.warn('error parsing typedData', e)
```
pkieltyka, line 383: how about updating the line to `console.warn('walletRequestHandler: error parsing typedData', e)` just to give more context?

packages/provider/src/utils.ts
```diff
@@ -23,6 +23,46 @@ export const prefixEIP191Message = (message: BytesLike): Uint8Array => {
   }
 }
 
+export const removeEIP191Prefix = (prefixedMessage: Uint8Array): Uint8Array => {
```
pkieltyka, line 26: lets rename this to `trimEIP191Prefix` .. minor thing, but I think "trim" is more common word, which reminds me of like "trimming whitespace"

pkieltyka (commented):
LGTM, tiny suggestions then lets merge

packages/provider/src/utils.ts
pkieltyka: I've updated the PR :)

PR #383, opened 2023-06-28
Avoid wallet deploy on v2 sdk

Agusx1211:
- [ ] Default `signMessage` and `signTypedMessage` to `sequenceVerified = true`. This means that these signatures won't trigger a wallet deployment.
- [ ] Pass `authorizeVersion == 2` to the connect options, this way we don't deploy the wallet upon connect (polygon/mumbai)


Agusx1211:
don't merge: I found an issue with the migrated wallets, `signMessage` doesn't validate correctly (the onchain contract can't validate a v2 signature, and the counterfactual validation doesn't match the v2 context)

Agusx1211:
Fixed it by still requiring the wallet deployment on `sequenceVerified == true` if the wallet is not deployed but the initial version wasn't 2

PR #388, opened 2023-07-04
V2 prettier format

corbanbrook:
Noticed that sequence.js hasnt been prettier formatted in a while. 

Not sure if we want to take a PR like this now against the v2 branch, or wait until we cut the official v2.0.0 release and merge this into master before doing a prettier fix


corbanbrook:
Closing for now... i think its better to wait

PR #397, opened 2023-07-19
Export network constants

corbanbrook:
Moves ChainId, NetworkId and networks (NetworkMetadata) to constants.ts which is also exported as @0xsequence/network/constants

NetworkConfig type is now an intersection of NetworkMetadata constant and configurable options. 

I believe this is a good code organization win as well as allows for importing constant metadata into other projects like design-system which needs chainId and Network details for NetworkTag components and others without pulling in a dependency on ethers.js

### DIFF
--- a/packages/network/src/config.ts
+++ b/packages/network/src/config.ts
@@ -3,60 +3,7 @@ import { Indexer } from '@0xsequence/indexer'
 import { Relayer, RpcRelayerOptions } from '@0xsequence/relayer'
 import { findNetworkConfig, stringTemplate, validateAndSortNetworks } from './utils'
 import { isBigNumberish } from '@0xsequence/utils'
-
-export enum ChainId {
-  // Ethereum
-  MAINNET = 1,
-  ROPSTEN = 3,
-  RINKEBY = 4,
-  GOERLI = 5,
-  KOVAN = 42,
-
-  // Polygon
-  POLYGON = 137,
-  POLYGON_MUMBAI = 80001,
-  POLYGON_ZKEVM = 1101,
-
-  // BSC
-  BSC = 56,
-  BSC_TESTNET = 97,
-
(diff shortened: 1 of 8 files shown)


packages/network/src/config.ts
```diff
@@ -366,22 +58,24 @@ const genUrls = (network: string) => {
...
     indexerUrl: indexerURL(network)
   }
 }
 
 export const allNetworks = validateAndSortNetworks([
+  {
+    ...networks[ChainId.POLYGON],
+    ...genUrls('polygon'),
+    isDefaultChain: true,
+    isAuthChain: true
```
pkieltyka, line 73: I think we can remove `isAuthChain` in general now with this version..? but, we can perhaps make a separate ticket and remove it separately to keep this PR just on the moving to constants, etc

packages/network/src/constants.ts
```diff
@@ -0,0 +1,339 @@
...
+  chainId: ChainId
+  type?: NetworkType
+  name: string
+  title?: string
+  blockExplorer?: {
+    name: string
+    rootUrl: string
+  }
+  ensAddress?: string
+  testnet?: boolean // Deprecated
```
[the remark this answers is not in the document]
corbanbrook: We still use it for compatibility. But are moving towards `type: NetworkType`

packages/network/src/config.ts
attente: Yeah, unfortunately I think we have to keep this because v1 sequence.js expects this property and at some point the wallet passes these networks to the dapp.

pkieltyka: makes sense, lets keep it for now, and later we can abstract it in another way

Agusx1211 (approved):

attente (approved):

PR #399, opened 2023-07-20
Re-implement sequence providers

Agusx1211:
- Implements `SequenceClient`, connects to sequence directly
- Remove the middleware stack when connecting to wallet-webapp directly
- Replaces the `Web3Provider` with a proper `SequenceProvider` that extends `BaseProvider`
- `SequenceProvider` supports network switching
- Removed redundant `Wallet` class, `provider` takes its place
- `SequenceSigner` implements an `ethers.Signer` for wallet-webapp
- Add single source of truth for default chain id (`SequenceClient`)
image


packages/provider/src/utils.ts
```diff
@@ -111,8 +111,8 @@ export const isWalletUpToDate = (status: AccountStatus): boolean => {
 }
 
 export interface ItemStore {
-  getItem(key: string): Promise<string | null>
```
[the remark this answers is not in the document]
Agusx1211: This store is used dapp-side, so I figure 100% of integrations are using the default (local storage). Making this async is a bit problematic, because ethers expects the provider to know the "initial network" in a sync way.

I say it's better to leave it like this for now, and if we want to support different stores later we can revisit it.

packages/provider/src/client.ts
```diff
@@ -0,0 +1,380 @@
...
+    return this.store.removeItem(SequenceClientSession.SESSION_LOCALSTORE_KEY)
+  }
+}
+
+/**
+ *  The wallet webapp doesn't really care what's the "default chain" for the user.
+ *  so we don't even bother to send this information to the wallet. Instead, we
+ *  track it locally using storage, that way the data stays always in sync.
+ */
+export class DefaultChainIDTracker {
```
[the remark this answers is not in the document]
Agusx1211: This is true, and to be honest I don't have a very good argument for it.

My weak reasoning is that it makes it a bit more clear some of the inner logic of these elements, for example that `defaultChainId` has an initial value that is independent of whatever is defined on the store, or that the session determines if it's connected or not using its internal state.

Agusx1211:
New docs for this:

https://github.com/0xsequence/docs/pull/111

packages/0xsequence/tests/browser/wallet-provider/dapp.test.ts
```diff
@@ -301,11 +298,11 @@ export const tests = async () => {
 
       // Record wallet deployed state before, so we can check the receipt.to below. We have to do this
       // because a wallet will automatically get bundled for deployment when it sends a transaction.
-      const beforeWalletDeployed = await wallet.isDeployed()
+      const beforeWalletDeployed = await hardhatProvider.getCode(wallet.getAddress()) !== '0x'
```
pkieltyka, line 301: why not keep wallet.isDeployed() method..? its more convienent and better dev experience then `x.getCode(wallet.getAddress()) !== '0x'`

Agusx1211: The general rationale is that a dapp should never need to "care" about the wallet being deployed or not, this should always be a concern of the wallet only. With EIP-6492 this is even more clear now.

Same with the wallet context / config.

Agusx1211: I feel like having these methods add a sort of noise when a developer is working with our wallet provider. They may see them and wonder if they need to account for these scenarios or not, and the correct answer is that they don't.

PR #404, opened 2023-07-27
provider: reject toxic transaction requests

attente:
Fixes https://github.com/0xsequence/issue-tracker/issues/629.

### DIFF
--- a/packages/provider/src/transactions.ts
+++ b/packages/provider/src/transactions.ts
@@ -12,14 +12,7 @@ function validateTransaction(wallet: string, transaction: commons.transaction.Tr
     if (transaction.data) {
       const data = ethers.utils.arrayify(transaction.data)
       if (data.length >= 4) {
-        const selector = ethers.utils.hexlify(data.slice(0, 4))
-        switch (selector) {
-          case '0x7a9a1628': // execute((bool,bool,uint256,address,uint256,bytes)[],uint256,bytes)
-          case '0x61c2926c': // selfExecute((bool,bool,uint256,address,uint256,bytes)[])
-            break
-          default:
-            throw new Error('self calls other than execute and selfExecute are forbidden')
-        }
+        throw new Error('self calls are forbidden')
       }
     }
   }
--- a/packages/provider/src/transactions.ts
+++ b/packages/provider/src/transactions.ts
@@ -1,0 +1,30 @@
+import { commons } from '@0xsequence/core'
+import { ethers } from 'ethers'
+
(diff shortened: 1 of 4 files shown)


packages/provider/src/transactions.ts
```diff
@@ -0,0 +1,30 @@
...
+import { ethers } from 'ethers'
+
+export function validateTransactionRequest(wallet: string, transaction: commons.transaction.Transactionish) {
+  const transactions = commons.transaction.fromTransactionish(wallet, transaction)
+  const unwound = commons.transaction.unwind(wallet, transactions)
+  unwound.forEach(transaction => validateTransaction(wallet, transaction))
+}
+
+function validateTransaction(wallet: string, transaction: commons.transaction.Transaction) {
+  if (transaction.to.toLowerCase() === wallet.toLowerCase()) {
```
Agusx1211, line 11: ALL calls to self should fail, there is no reason for a dapp to be asking the wallet to call itself. If this is a problem we could get a bit more fancy, but for now I think this would be wise.

packages/provider/src/transports/wallet-request-handler.ts
```diff
@@ -471,12 +474,14 @@ export class WalletRequestHandler implements ExternalProvider, JsonRpcHandler, P
           // and would have prompted the user upon signing.
 
           // https://eth.wiki/json-rpc/API#eth_sendRawTransaction
-          if (commons.transaction.isSignedTransactionBundle(request.params![0])) {
-            const txChainId = BigNumber.from(request.params![0].chainId).toNumber()
-            const tx = await account.relayer(txChainId)!.relay(request.params![0])
+          const transaction = request.params![0]
+          if (commons.transaction.isSignedTransactionBundle(transaction)) {
+            validateTransactionRequest(account.address, transaction.transactions)
```
Agusx1211, line 479: I don't think is flow is even properly supported, but just in case this should happen on the sign, not on the send. After the transaction is signed is already too late.

packages/provider/src/transactions.ts
attente: I'm not sure if this is still the case, but we needed to be able to wrap some transactions in a `selfExecute` bundle in order to ensure the relayer fee was still getting paid while the internal bundle was treated as atomic:

https://github.com/horizon-games/issue-tracker/issues/4647#issuecomment-853226890

Agusx1211: I think we don't wan to expose this to dapps, gas refunds are handled entirely by wallet-webapp.
Besides, priority should be safety.

Agusx1211 (approved):

PR #411, opened 2023-08-04
Mutate connectDetails.session.networks isDefaultChain to return the dapps connectOptions.networkId instead

corbanbrook:
When dapps initWallet they specify their own default chain id, however when they connect to the wallet, the wallet passes its own default chain id settings to the dapp. This causes problems!

When dapps connect they receive connectDetails back which includes the session from the wallet. This session contains networks of `NetworkConfig[]` with the `isDefaultChain: true` of the wallets default chain. This causes the dapp to override its own specified default network with the wallets.

If the dapp window is reloaded it is already in a connected state so it doesnt need to connect again and uses its own default chain. This fix is to ensure the dapp always stays on its own default chain even after a connect.

During connect we check to see if the dapp passes in connectOptions.networkId and if one is specified we mutate the wallets session.networks to change the returned defaultId.

### DIFF
--- a/packages/provider/src/transports/wallet-request-handler.ts
+++ b/packages/provider/src/transports/wallet-request-handler.ts
@@ -3,6 +3,7 @@ import { signAuthorization, AuthorizationOptions } from '@0xsequence/auth'
 import { commons } from '@0xsequence/core'
 import {
   ChainId,
+  findNetworkConfig,
   findSupportedNetwork,
   JsonRpcHandler,
   JsonRpcRequest,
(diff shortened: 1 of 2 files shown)


packages/provider/src/transports/wallet-request-handler.ts
```diff
@@ -200,6 +201,18 @@ export class WalletRequestHandler implements ExternalProvider, JsonRpcHandler, P
     const connectDetails: ConnectDetails = promptConnectDetails
     if (connectDetails.connected && !connectDetails.session) {
       connectDetails.session = await this.walletSession()
+
+      if (options?.networkId) {
+        const network = findNetworkConfig(connectDetails.session?.networks || [], options.networkId)
```
attente, line 206: Maybe we can change to:

```
if (options?.networkId && connectDetails.session?.networks?.length > 0) {
  const network = findNetworkConfig(connectDetails.session.networks, options.networkId)
```
```

attente (approved):

tolgahan-arikan (approved):

PR #417, opened 2023-08-16
Pass client version with connect options

corbanbrook:


packages/provider/src/client.ts
```diff
@@ -19,6 +19,8 @@ import { TypedData } from '@0xsequence/utils'
 import { toExtended } from './extended'
 import { ethers } from 'ethers'
 
+import packageJson from '../package.json'
```
corbanbrook, line 22: Im not sure this is the best way to get access to the version. importing the entire packageJson is wasteful. 

Unfamiliar with preconstruct but normally if i had access to a build script like in webpack I would import package.json within the build script and then just define a global to be picked up within the lib. like process.env.CLIENT_VERSION.   Any ideas?

Agusx1211 (approved):
LGTM

PR #428, opened 2023-09-28
WalletRequestHandler.notifyChainChanged

corbanbrook:
Adding notifyChainChanged to wallet request handler to allow broadcasting connected site chain changes to origin.

### DIFF
--- a/packages/provider/src/transports/wallet-request-handler.ts
+++ b/packages/provider/src/transports/wallet-request-handler.ts
@@ -771,13 +771,18 @@ export class WalletRequestHandler implements ExternalProvider, JsonRpcHandler, P
     this.events.emit('disconnect')
   }
 
+  notifyChainChanged(chainId: number, origin: string) {
+    console.log('-- notifyChainChanged', chainId, origin)
+    this.events.emit('chainChanged', ethers.utils.hexValue(chainId), origin)
+  }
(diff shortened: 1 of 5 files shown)


packages/provider/src/transports/wallet-request-handler.ts
```diff
@@ -771,13 +771,18 @@ export class WalletRequestHandler implements ExternalProvider, JsonRpcHandler, P
     this.events.emit('disconnect')
   }
 
+  notifyChainChanged(chainId: number, origin: string) {
+    console.log('-- notifyChainChanged', chainId, origin)
```
attente, line 775: Remove this log.

packages/provider/src/transports/wallet-request-handler.ts
```diff
@@ -771,13 +771,18 @@ export class WalletRequestHandler implements ExternalProvider, JsonRpcHandler, P
     this.events.emit('disconnect')
   }
 
+  notifyChainChanged(chainId: number, origin: string) {
```
attente, line 774: `origin` is ok to be mandatory here, right? I think that's fine, just wanted to check with you.

attente (changes requested):

attente (approved):

PR #439, opened 2023-11-07
project access keys

pkieltyka:


packages/indexer/src/index.ts
```diff
@@ -26,13 +26,16 @@ export enum SequenceIndexerServices {
...
+export class SequenceIndexer extends IndexerRpc {
   constructor(
     hostname: string,
-    public authorization?: {
-      jwtAuth?: string
-      accessKey?: string
-    }
+    public projectAccessKey?: string,
+    public jwtAuth?: string
+    // public authorization?: {
```
pkieltyka, line 35: perhaps using .. `new SequenceIndexer('https://polygon-indexer.sequence.app', { projectAccessKey: 'etcetc' })` is better dev experience, I believe so.

packages/indexer/src/index.ts
```diff
@@ -2,55 +2,32 @@ export * from './indexer.gen'
...
-  OPTIMISM = 'https://optimism-indexer.sequence.app',
-  AVALANCHE = 'https://avalanche-indexer.sequence.app',
-  GNOSIS = 'https://gnosis-indexer.sequence.app',
-
-  BSC = 'https://bsc-indexer.sequence.app',
-  BSC_TESTNET = 'https://bsc-testnet-indexer.sequence.app',
-
-  GOERLI = 'https://goerli-indexer.sequence.app'
-}
-
```
tolgahan-arikan, line 26: I removed these and SequenceIndexerServices below as they seem unnecessary and aren't used in webapp. I asked Deno if it's used on builder or marketplace. Will put them back in case they are used there.

tolgahan-arikan: update: Deno confirmed they are not used in builder or marketplace, so it should be ok

PR #447, opened 2023-11-29
WaaS ethers wrapper

ScreamingHawk:


packages/waas-ethers/src/signer.ts
```diff
@@ -0,0 +1,127 @@
...
+  async getAddress(): Promise<string> {
+    return this.sequence.getAddress()
+  }
+
+  // Ensure that sequence and the provider are on matching networks
+  private async _ensureNetworkMatch(providerRequired: boolean): Promise<void> {
+    if (providerRequired && !this.provider) {
+      throw new Error('Provider is required')
+    }
+    if (this.provider && (await this.provider.getNetwork()).chainId !== networks.toNetworkID(this.sequence.getNetwork())) {
```
Agusx1211, line 21: Sequence Auth supports multiple networks, maybe is it worth to accept any provider as long as the network is among the supported ones by Sequence Auth?

packages/waas-ethers/src/signer.ts
```diff
@@ -0,0 +1,127 @@
+import { BigNumber, ethers } from 'ethers'
+import { CommonAuthArgs, Sequence, networks } from '@0xsequence/waas'
+
+export class SequenceSigner extends ethers.Signer {
+  constructor(
+    private readonly sequence: Sequence,
```
Agusx1211, line 6: Maybe is it worth to have either an static method or the option to pass the sequence options directly here? so the wrapping doesn't have to be done manually

Agusx1211 (changes requested):
LGTM, small changes

PR #452, opened 2023-12-08
upgrade deps

pkieltyka:
upgrades all deps, other then ethers which remains at ethers@v5 for time being


packages/account/tests/account.spec.ts
```diff
@@ -88,7 +88,7 @@ describe('Account', () => {
   }
 
   before(async () => {
-    provider1 = new ethers.providers.Web3Provider(hardhat.network.provider.send)
+    provider1 = new ethers.providers.Web3Provider(hardhat.network.provider as any)
```
pkieltyka, line 91: this was a necessary change, otherwise caused ethers to detect the network, and the typings are different between hardhat.network.provider and our receiver

PR #453, opened 2023-12-08
provider analytics for projects

pkieltyka:
this PR adds support to project analytics support sequence.js via a new @databeat/tracker. 

1. it adds analytics.ts which passes projectAccessKey to setup databeat/tracker and uses new custom header option to associate events with a project
2. creates new `this.analytics` instance which will automatically track loads on instantiation (if analytics is enabled on provider that is), but also allows to track custom events as well

### DIFF
--- a/packages/provider/src/client.ts
+++ b/packages/provider/src/client.ts
@@ -150,7 +150,7 @@ export class SequenceClient {
       throw new Error('Invalid transport')
     }
 
-    const defaultChainId = options?.defaultChainId ?? 1
+    const defaultChainId = options?.defaultChainId
     this.defaultEIP6492 = options?.defaultEIP6492 ?? false
 
     this.session = new SequenceClientSession(store)


packages/provider/package.json
```diff
@@ -29,6 +29,7 @@
     "@0xsequence/relayer": "workspace:*",
     "@0xsequence/utils": "workspace:*",
     "@0xsequence/wallet": "workspace:*",
+    "@databeat/tracker": "^0.7.0",
```
pkieltyka, line 32: note, we'll need to release databeat/tracker 0.8.0 first

taylanpince:
LGTM – @pkieltyka or @attente If this is good for you as well, we would like to merge and release as 1.8.0 tomorrow. We need to ensure node-gateway proxy is fully deployed beforehand.

packages/provider/src/client.ts
```diff
@@ -135,13 +138,10 @@ export class SequenceClient {
...
+  public readonly projectAccessKey?: string
+  public readonly analytics?: Analytics
 
-  constructor(
-    transport: ProviderTransport | MuxTransportTemplate,
-    store: ItemStore,
-    options?: SequenceClientOptions | number,
-    public projectAccessKey?: string
-  ) {
+  constructor(transport: ProviderTransport | MuxTransportTemplate, store: ItemStore, options?: SequenceClientOptions) {
```
attente, line 144: If `options` is no longer accepted as a `number`, then there are code paths that are no longer required.

packages/provider/src/analytics.ts
```diff
@@ -0,0 +1,50 @@
...
+export class Analytics extends Databeat<EventTypes> {}
+
+// Setup analytics tracker
+export const setupAnalytics = (projectAccessKey: string, server?: string) => {
+  if (!server) {
+    server = 'https://nodes.sequence.app'
+  }
+
+  // disable tracking if projectAccessKey is not set
+  const noop = !(projectAccessKey && projectAccessKey.length > 0)
```
attente, line 27: Just `const noop = !projectAccessKey` is fine because `''` is falsy and non-empty strings are truthy.

packages/provider/src/analytics.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  // auth
+  const auth: Auth = {}
+  if (projectAccessKey && projectAccessKey.length > 0) {
```
attente, line 31: Again, just `if (projectAccessKey)` is fine because `''` is falsy and non-empty strings are truthy.

packages/provider/src/analytics.ts
```diff
@@ -0,0 +1,50 @@
...
+export type EventTypes = keyof typeof EventType
+export type Event = DatabeatEvent<EventTypes>
+
+// Analytics sub-class to add some custom helper methods
+export class Analytics extends Databeat<EventTypes> {}
+
+// Setup analytics tracker
+export const setupAnalytics = (projectAccessKey: string, server?: string) => {
+  if (!server) {
+    server = 'https://nodes.sequence.app'
```
attente, line 23: Shouldn't this be the analytics server where databeat events are tracked?

packages/provider/src/analytics.ts
```diff
@@ -0,0 +1,50 @@
...
+    defaultEnabled: true,
+    privacy: { userIdHash: true, userAgentSalt: false },
+    initProps: () => {
+      if (!isBrowser()) {
+        return {}
+      } else {
+        const origin = window.location.origin
+        return {
+          origin
+        }
```
attente, line 46: Minor nit, but can we just instead write:

```
return { origin: window.location.origin }
```

attente (changes requested):

packages/provider/src/analytics.ts
tolgahan-arikan: afaik node-gateway proxy will handle it

attente (approved):

PR #457, opened 2023-12-12
Waas

marino39:


packages/waas/src/auth.ts
```diff
@@ -0,0 +1,496 @@
...
+
+    const key = await window.crypto.subtle.importKey('raw', decodeHex(plainHex), cbcParams, false, ['encrypt'])
+    const payloadBytes = new TextEncoder().encode(JSON.stringify(payload))
+    const encrypted = await window.crypto.subtle.encrypt(cbcParams, key, payloadBytes)
+    const payloadCiphertext = encodeHex(new Uint8Array([...cbcParams.iv, ...new Uint8Array(encrypted)]))
+    const payloadSig = await this.waas.signUsingSessionKey(payloadBytes)
+
+    return {
+      headers: {
+        'X-Sequence-Tenant': this.config.tenant
```
pkieltyka, line 258: lets update this to `X-Access-Key`

packages/waas/src/networks.ts
```diff
@@ -0,0 +1,60 @@
+
+const RPC_BASE = 'https://nodes.sequence.app/'
+
+const nameToId = {
+  'mainnet': 1,
```
pkieltyka, line 5: we should be using `@0xsequence/network` instead, it exports a `networks` object which is our canonical network list of names and ids

taylanpince: @marino39 Can you fix this? As Peter mentioned above, `networks` package offers a really nice interface for this already and there is no need to hardcode these values here. Let's standardize.

taylanpince (changes requested):

packages/waas/src/networks.ts
marino39: Yes, I have missed that one.

PR #458, opened 2023-12-14
Access key usage for singleSigner

tolgahan-arikan:


packages/auth/src/session.ts
```diff
@@ -180,7 +187,8 @@ export class Session {
         tracker,
         networks,
         contexts,
-        orchestrator
+        orchestrator,
+        projectAccessKey
```
attente, line 191: Technically there is another place below where this should be added in the new account case, do you mind adding it there as well? Line 279

attente (changes requested):

PR #463, opened 2024-01-11
waas chain list

marino39:


Agusx1211:
LGTM but let me see what is going on with the CI

PR #475, opened 2024-01-30
Pass session address as nonce to identity providers

patrislav:


packages/waas/src/base.ts
```diff
@@ -201,13 +197,17 @@ export class SequenceWaaSBase {
...
 
-    const result = await openSession({ proof, lifespan: DEFAULT_LIFESPAN })
+    const packet = await openSession({
+      signer: await this.getSigner(),
+      proof,
+      lifespan: DEFAULT_LIFESPAN,
+    })
 
-    await Promise.all([this.status.set('pending'), this.signer.set(result.signer.privateKey)])
+    await Promise.all([this.status.set('pending')])
```
Agusx1211, line 206: If this is a single, there is no need to use `Promise.all`

packages/waas/src/email.ts
```diff
@@ -77,12 +77,15 @@ export class EmailAuth {
...
 
-  public async finalizeAuth({ instance, email, answer }: { instance: string, email: string, answer: string }): Promise<Identity> {
+  public async finalizeAuth({ instance, email, answer, sessionAddress }: { instance: string, email: string, answer: string, sessionAddress?: string }): Promise<Identity> {
     const res = await this.cognito().send(new RespondToAuthChallengeCommand({
       ClientId: this.clientId,
       Session: instance,
       ChallengeName: 'CUSTOM_CHALLENGE',
       ChallengeResponses: { USERNAME: email, ANSWER: answer },
+      ClientMetadata: {
+        sessionAddress: sessionAddress || "",
```
Agusx1211, line 87: Should ever sessionAddress be null here?

Agusx1211:
LGTM @taylanpince , just two small comments

patrislav:
Thanks, updated

PR #477, opened 2024-01-31
update rpc contract with latest changes

LukasJenicek:


corbanbrook (approved):

packages/api/src/api.gen.ts
```diff
@@ -396,6 +422,7 @@ export interface SendPasswordlessLinkReturn {
   status: boolean
 }
 export interface FriendListArgs {
+  nickname: string
```
corbanbrook, line 425: nickname should not be a required property

LukasJenicek: Passing an empty string will work.

corbanbrook: Thats what im doing for now, but we should correct this

pkieltyka: yes, change the RIDL to make it optional @LukasJenicek

LukasJenicek: Totally forgot about this one. It's merged  `stack/api` :https://github.com/0xsequence/stack/pull/774 and deployed: https://buildkite.com/horizon/sequence-stack-build/builds/1068

LukasJenicek: I just did not like the idea that we have to reference string value by pointer, but I think there is nothing I can do about it

PR #478, opened 2024-01-31
waas: New Intents & SECP256R1

marino39:


packages/waas/src/payloads/signers/index.ts
```diff
@@ -0,0 +1,28 @@
+import { newSECP256K1RandomSigner, newSECP256K1Signer } from "./secp256k1";
+import { newSECP256R1Signer, newSECP256R1RandomSigner } from "./secp256r1";
+
+export type PayloadSigner = {
```
pkieltyka, line 4: maybe rename it to, SessionSigner ..? its a bit more clear

packages/waas/src/payloads/signers/secp256r1.ts
```diff
@@ -0,0 +1,61 @@
...
+export async function newSECP256R1Signer(privateKey: string): Promise<PayloadSigner> {
+  const [pubKey, privKey] = await loadSECP256R1Keys(privateKey)
+
+  const exPrivateKey = await window.crypto.subtle.exportKey('jwk', privKey)
+  const exPublicKey = await window.crypto.subtle.exportKey('jwk', pubKey)
+  const keys = [exPublicKey, exPrivateKey]
+
+  const encoder = new TextEncoder()
+  return {
+    privateKey: ethers.utils.hexlify(encoder.encode(JSON.stringify(keys))),
```
pkieltyka, line 13: btw, the advantage of the r1 curve is we can great a key in window.crypto that is not exportable, and never gain access to the private key. This is what makes it more secure. Therefore, the privateKey wouldn't be available here. But you should return the hex of the public key.

Lmk what you thoughts are though...?

packages/waas/src/payloads/signers/secp256r1.ts
```diff
@@ -0,0 +1,61 @@
...
+  }
+}
+
+export async function newSECP256R1RandomSigner(): Promise<PayloadSigner> {
+  const generatedKeys = await window.crypto.subtle.generateKey(
+    {
+      name: "ECDSA",
+      namedCurve: "P-256",
+    },
+    true,
```
pkieltyka, line 39: should be false... `extractable` argument.. so the private key is not extractable.

FYI, the browser will persist the private key for you between sessions, etc. so it will be there. Just need to check the APIs etc. I believe it stored in IndexedDB. If you're not sure, you can ping Corban and ask him, he's done some window.crypto stuff before.

packages/waas/src/payloads/signers/index.ts
```diff
@@ -0,0 +1,28 @@
...
+  privateKey: string
```
pkieltyka, line 5: lets update the type to have `privateKey?: string` but add `publicKey: string` for k1 curve, you can also get the public key from ethers, or we can just return the address generated. For r1 curve, lets return the public key in hex.

packages/waas/src/payloads/signers/index.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  verifier(): Promise<string>
+  signMessage(message: string | Uint8Array): Promise<string>
+}
```
pkieltyka, line 9: I think we'll also want to add the method `verify(message: string, signature: string): Promise<boolean>` or similar, to complete this interface of signing and verifying for a pub/priv signing algorithm

packages/waas/src/payloads/signers/index.ts
```diff
@@ -0,0 +1,28 @@
+import { newSECP256K1RandomSigner, newSECP256K1Signer } from "./secp256k1";
+import { newSECP256R1Signer, newSECP256R1RandomSigner } from "./secp256r1";
+
+export type PayloadSigner = {
+  privateKey: string
+
+  verifier(): Promise<string>
```
pkieltyka, line 7: I propose, rename `verifier(): Promise<string>` to just `keyType(): string` and have it return "secp256k1" or "secp256r1" or instead of "keyType()" name, use "alg()" aka algorithm, similar to what jwt calls it

packages/waas/src/base.ts
```diff
@@ -34,7 +33,8 @@ import { SimpleNetwork, WithSimpleNetwork, toNetworkID } from './networks'
 type status = 'pending' | 'signed-in' | 'signed-out'
 
 const SEQUENCE_WAAS_WALLET_KEY = '@0xsequence.waas.wallet'
-const SEQUENCE_WAAS_SIGNER_KEY = '@0xsequence.waas.signer'
+const SEQUENCE_WAAS_SESSION_ID_KEY = '@0xsequence.waas.session_id'
+const SEQUENCE_WAAS_PAYLOAD_SIGNER_KEY = '@0xsequence.waas.payload_signer'
```
pkieltyka, line 37: lets rename to `SEQUENCE_WAAS_SESSION_SIGNER_KEY = '@0xsequence.waas.session_signer'` .. much more clear

packages/waas/src/payloads/signers/secp256r1.ts
pkieltyka: for the r1 publicKey(): Promise<string> method, we can use `ethers.utils.hexlify( window.crypto.subtle.exportKey('raw', pubKey))` which will return 130 character length string in hex of the 65 byte public key.

potentially to.... maybe we should add `address: Promise<string>` to the SessionSigner interface, for both k1 and r1 of course.. where k1 already has address, and for r1 address we could do like `'0x'+publicKey().slice(0,40)`

and we have the `alg()` to tell us which curve it is..

in terms of "verify()" if we have just the alg, address, message and signature... I

marino39: It uses export raw for ``publicKey``.  I don't like the name ``sessionAddress`` as the address makes sense only in ethereum. So I would rather use names such as ``sessionId``, ``sessionKeyType``, ``sessionPublicKey``, ``signature``.

packages/waas/src/payloads/signers/secp256r1.ts
marino39: I couldn't find a way to persist the non-exportable key. Thank you for pointing me to this solution. I already have found a couple of examples.

packages/waas/src/payloads/signers/index.ts
marino39: At the moment this field is used to persist the key. Not sure how this will look like when we will be using ``IndexedDB``. I guess it's to be determined.

packages/waas/src/payloads/signers/index.ts
marino39: I did not implement the verify method because we never will be using this in this context. It might be useful for unit tests though. So I guess I will do that.

packages/waas/src/payloads/signers/index.ts
marino39: Yes, the SessionSigner interface should look along these lines:
```ts
export type SessionSigner = {
    keyType(): string // the alg
    publicKey(): string // the public key for verification
    sign(...): string // the signing function
    verify(): string // the verification function
}
```

packages/waas/src/base.ts
marino39: I think this will be removed when we use the ``IndexedDB``. I will apply if not.

packages/waas/src/payloads/signers/secp256r1.ts
pkieltyka: Cool.. were you able to find a way to persist the non-exportable key..?

because, tbh, if we can't use the non-exportable key, then there is no point in adding R1 support at all, and we should just skip it. The only value to r1 is we can have non-exportable private keys.

packages/waas/src/payloads/signers/index.ts
pkieltyka: yes, its easy to do, and for completion we should do it

packages/waas/src/payloads/signers/secp256r1.ts
pkieltyka: agreed, lets go with the name `sessionId`

packages/waas/src/payloads/signers/secp256r1.ts
marino39: This is what I have found:
https://gist.github.com/saulshanabrook/b74984677bccd08b028b30d9968623f5

They claim that you can store it in IndexedDB as non-exportable object. However they are not sure if it's more secure.

packages/waas/src/defaults.ts
```diff
@@ -7,7 +7,6 @@
 export const TEST = {
   rpcServer: 'https://d14tu8valot5m0.cloudfront.net',
   kmsRegion: 'us-east-2',
```
pkieltyka, line 9: do we still use `kms` for anything..?

pkieltyka: dont we still need `idpRegion` .. no?

packages/waas/src/session/secp256k1.ts
```diff
@@ -0,0 +1,54 @@
+import { ethers } from "ethers";
+import { openDB } from "idb";
+import { Session } from "./index";
+
+const idbName = 'waas-session-p256k1'
```
pkieltyka, line 5: I think call it ... `seq-waas-session-p256k1` ... etc.... add the `seq-` prefix.. so it doesnt conflict with anything in anyway.. etc..

packages/waas/src/session/secp256r1.ts
```diff
@@ -0,0 +1,86 @@
+import { ethers } from "ethers";
+import { Session } from "./index";
+import { KeyTypes } from './keyTypes'
+
+import { openDB } from 'idb';
+
+const idbName = 'waas-session-p256r1'
```
pkieltyka, line 7: lets add `seq-` prefix to the names

packages/waas/src/session/secp256r1.ts
```diff
@@ -0,0 +1,86 @@
...
+          message = ethers.utils.arrayify(message)
+        } else {
+          message = encoder.encode(message)
+        }
+      }
+      const signatureBuff = await window.crypto.subtle.sign({name: 'ECDSA', hash: {name: 'SHA-256'}}, keys.privateKey, message)
+      return ethers.utils.hexlify(new Uint8Array(signatureBuff))
+    },
+    clear: async () => {
+      await db.delete('waas-session', sessionId)
```
pkieltyka, line 42: should use `idbStoreName`

packages/waas/package.json
```diff
@@ -18,6 +18,7 @@
     "@aws-sdk/client-cognito-identity-provider": "^3.445.0",
     "@aws-sdk/client-kms": "^3.429.0",
```
pkieltyka, line 19: can we remove `@aws-sdk/client-kms` or any of the aws packages now that we've removed the transport key..? just noting we should purge any unused deps from here.

pkieltyka (commented):
very very nice!! lets some comments

packages/waas/src/defaults.ts
patrislav: yeah, we don't need either, removed

PR #483, opened 2024-02-15
Ethers v6

corbanbrook:


packages/0xsequence/tests/browser/mock-wallet/mock-wallet.test.ts
```diff
@@ -42,15 +42,15 @@ const main = async () => {
     {
       name: 'hardhat',
       chainId: 31337,
-      rpcUrl: provider.connection.url,
+      rpcUrl: provider._getConnection().url,
```
pkieltyka, line 45: this is kinda weird.. we shouldn't need to use `_` methods.. is there nothing else with this info on the object..?

packages/account/src/account.ts
```diff
@@ -82,11 +82,11 @@ class Chain0Reader implements commons.reader.Reader {
     return undefined
   }
 
-  async nonce(_wallet: string, _space: ethers.BigNumberish): Promise<ethers.BigNumberish> {
-    return ethers.constants.Zero
+  async nonce(_wallet: string, _space: BigIntish): Promise<bigint> {
```
pkieltyka, line 85: ethers still has the `ethers.BigNumberish` type, but now it uses native bigInt or strings ..

Our response type is now `Promise<bigint>` ... what is `bigint` ..? Im just wondering what is most sensible return type..

PR #488, opened 2024-02-28
waas: sequence.feeOptions

marino39:
demo-waas-auth: https://github.com/0xsequence/demo-waas-auth/pull/15


packages/waas/src/intents/transactions.ts
```diff
@@ -83,10 +122,54 @@ export function sendTransactions({
         value: ethers.BigNumber.from(tx.value || 0).toHexString(),
         data: ethers.utils.hexlify(tx.data || [])
       }
-    })
+    }),
+    transactionsFeeQuote
   })
 }
 
+function withTransactionFee(transactions: Transaction[], feeOption?: FeeOption): Transaction[] {
```
patrislav, line 130: This function will mutate the `transactions` array passed into `sendTransactions` - possibly unexpectedly by the caller. Is this expected?

Otherwise you could create a copy:
```
const extendedTxns = [...transactions]
// ...
extendedTxns.push(/* ... */)
// ...
return extendedTxns
```

packages/waas/src/intents/transactions.ts
```diff
@@ -104,6 +187,31 @@ export function getTransactionReceipt({
...
+  return erc1155({ tokenAddress: token, to, vals, data })
+}
+
+export function sendDelayedEncodeArgsToTransaction({ to, value, abi, func, args, ...otherArgs }: Omit<SendDelayedEncodeArgs, 'chainId'>): Transaction {
+  return delayedEncode({
+    to,
+    value: ethers.BigNumber.from(value).toString(),
+    data: { abi, func, args }
+  })
+}
```
patrislav, line 213: These functions look quite unwieldy - the base `erc20`, `erc721`, etc. functions were supposed to be the ones used by library consumers to easily turn those into `Transaction`. Now we have another way to do it which is probably wrong...

Can we instead move this functionality to the simple `erc20` etc. functions? E.g. so that they support both of the formats and consumers aren't confused which to use.

marino39: Unfortunately, those functions were never public, their params are different(not only names but also types). In addition to that function names such as ``erc20``, ``erc1155``... do not really say much about what these functions do.

My first approach just broke SDK API contract but was clean. This one preserves API but is not as clean. We could alter ``erc20``... functions so they work with bot params but again it will not be very clean and intuitive to use.

@taylanpince @pkieltyka WDYT?

taylanpince: I prefer the cleaner method here. Nobody is using this part of the SDK on production yet, this is still a good time to make a breaking change. I vote for adding a breaking change and modifying the original methods.

taylanpince:
@marino39 I think this looks quite good, and the interface is way simpler now so going through the erc methods is not a big deal. I am good to go with this approach 👍

patrislav (approved):

PR #514, opened 2024-04-18
waas: fix toNetworkId function

patrislav:
This would fail if a string is passed to it that looks like a number.


marino39 (approved):
LGTM

PR #517, opened 2024-04-28
waas: add pluggable SubtleCryptoBackend

pkieltyka:


packages/waas/src/session/secp256r1.ts
```diff
@@ -1,13 +1,18 @@
...
 import { openDB } from 'idb'
 
 const idbName = 'seq-waas-session-p256r1'
 const idbStoreName = 'seq-waas-session'
 
-export async function newSECP256R1SessionFromSessionId(sessionId: string): Promise<Session> {
+// TODO: in order to support react-native we will have to create
+// an adapter for idb here for storage, and anywhere else 'idb' is used
+// in the entire project.
+
```
pkieltyka, line 14: Please read important note/TODO here if we want full RN support. This PR gets us towards it, but we need to add a similar backend for "idb" for a "Store" or "StoreBackend", etc. where on RN side we'd use like the mmkv thing, and in browsers we'd use idb, etc.

pkieltyka:
closing, as is included in https://github.com/0xsequence/sequence.js/pull/518

PR #518, opened 2024-04-28
Secure store backend for waas with keychain support

taylanpince:


packages/waas/src/base.ts
```diff
@@ -80,7 +82,9 @@ export class SequenceWaaSBase {
 
   constructor(
     public readonly config = { network: 1 } as SequenceBaseConfig,
-    private readonly store: Store = new LocalStore()
+    private readonly store: Store = new LocalStore(),
+    private readonly cryptoBackend: SubtleCryptoBackend | null = getDefaultSubtleCryptoBackend(),
+    private readonly secureStoreBackend: SecureStoreBackend
```
pkieltyka, line 87: I wonder if we should include a default for `secureStoreBackend` too..?

packages/waas/src/secure-store.ts
```diff
@@ -0,0 +1,59 @@
...
+export const getDefaultSecureStoreBackend = (): SecureStoreBackend | null => {
+    if (isIndexedDbAvailable()) {
+        return new IndexedDbSecureStoreBackend()
+    } else {
+        return null
+    }
+}
+
+export function isIndexedDbAvailable(): boolean {
+    return typeof indexedDB === 'object'
```
pkieltyka, line 18: btw, is it intentional to not use `window.indexedDB` here..? (just curious)

packages/waas/src/secure-store.ts
```diff
@@ -0,0 +1,59 @@
...
+
+export class IndexedDbSecureStoreBackend implements SecureStoreBackend {
+    constructor() {
+        if (!isIndexedDbAvailable()) {
+            throw new Error('IndexedDB is not available')
+        }
+    }
+
+    async get(dbName: string, dbStoreName: string, key: string): Promise<string | null> {
+        const db = await openDB(dbName)
```
pkieltyka, line 29: I don't think we probably want to open and close the db each time we get a value? maybe we can just add "open" and "close" methods to the SecureStoreBackend ..? and we can automatically call .open() in the constructor..? for close(), we can leave it to developers to call if they want. Just some ideas.. lmk what you think

... and, if you dont want to call .open() in the constructor for whatever reason, then add a private instance variable, and open if in the get/set/delete methods if private instance of db is undefined, etc..

packages/waas/src/secure-store.ts
```diff
@@ -0,0 +1,59 @@
...
+        const tx = db.transaction(dbStoreName, 'readonly')
+        const value = await db.get(dbStoreName, key)
+        await tx.done
+        db.close()
+        return value
+    }
+
+    async set(dbName: string, dbStoreName: string, key: string, value: string): Promise<boolean> {
+        const db = await openDB(dbName, 1, {
```
pkieltyka, line 38: lets move this to the open() method too. I think its safe to call upgrade() each time on open, the idea is it will create a new db and set the version to 1 if its never been created before according to https://github.com/jakearchibald/idb

packages/waas/src/session/secp256r1.ts
```diff
@@ -1,13 +1,20 @@
...
+import { SubtleCryptoBackend } from '../subtle-crypto'
 
 import { openDB } from 'idb'
+import { SecureStoreBackend } from '../secure-store'
 
 const idbName = 'seq-waas-session-p256r1'
 const idbStoreName = 'seq-waas-session'
 
-export async function newSECP256R1SessionFromSessionId(sessionId: string): Promise<Session> {
+// TODO: We need to update this to use the secure store backend
```
pkieltyka, line 12: if the CryptoKeyPair type becomes a problem, we could make the type `any` at least on our interface, but we'd need to typecast and confirm inside of the backend implementations.. in case we wanted to handle other kinds of types on the interface methods

packages/waas/src/base.ts
taylanpince: Since `secureStoreBackend` is required, I added a default in `auth.ts` and made sure it was null-checked before coming here.

packages/waas/src/secure-store.ts
taylanpince: Yes, intentional. This is how `idb` references it as well. Using `window.indexedDB` doesn't seem to work (at least when running tests)

packages/waas/src/secure-store.ts
taylanpince: Agreed – Previous implementation was only making this call when setting the key, but I think it's better to just do it on any open automatically. I'll change this.

packages/waas/src/secure-store.ts
taylanpince: Done, it now auto-handles db open on first call.

packages/waas/src/session/secp256r1.ts
taylanpince: I've moved to `any` type on the interface and converted R1 implementation to use SecureStoreBackend. KeychainStore won't work with R1 for now, but we can resolve that when R1 is ready for RN.

pkieltyka (approved):

PR #521, opened 2024-04-30
update pnpm lock version

pkieltyka:


pkieltyka:
replaced with https://github.com/0xsequence/sequence.js/pull/522

PR #524, opened 2024-05-03
Remove defaults from waas

taylanpince:
This is a **breaking change for implementors** since the constructor interface is changing. I think cleaning up like this is better but open to thoughts.

https://github.com/0xsequence/issue-tracker/issues/2346


patrislav (approved):
LGTM

PR #528, opened 2024-05-20
add SequenceIndexerOpts

xiam:
Reference: https://github.com/0xsequence/issue-tracker/issues/2358

### DIFF
--- a/packages/indexer/src/index.ts
+++ b/packages/indexer/src/index.ts
@@ -4,13 +4,40 @@ import { Indexer as IndexerRpc } from './indexer.gen'
 
 const fetch = typeof global === 'object' ? global.fetch : window.fetch
 
+export interface SequenceIndexerOpts {
+  JWTAuth?: string
+  ProjectAccessKey?: string
+}
+
 export class SequenceIndexer extends IndexerRpc {
+  public jwtAuth?: string
+  public projectAccessKey?: string
+
   constructor(
     hostname: string,
-    public projectAccessKey?: string,
-    public jwtAuth?: string
+    opts: SequenceIndexerOpts
+  );
+  constructor(
+    hostname: string,
+    projectAccessKey?: string,
+    jwtAuth?: string
+  );
+  constructor(
+    hostname: string,
+    ...args: any[]
   ) {
     super(hostname.endsWith('/') ? hostname.slice(0, -1) : hostname, fetch)
+
+    if (args.length === 1 && typeof args[0] === 'object') {
+      const opts = args[0] as SequenceIndexerOpts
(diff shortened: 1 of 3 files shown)


packages/indexer/src/index.ts
```diff
@@ -4,13 +4,40 @@ import { Indexer as IndexerRpc } from './indexer.gen'
...
     super(hostname.endsWith('/') ? hostname.slice(0, -1) : hostname, fetch)
+
+    if (args.length === 1 && typeof args[0] === 'object') {
+      const opts = args[0] as SequenceIndexerOpts
+      this.projectAccessKey = opts.ProjectAccessKey
+      this.jwtAuth = opts.JWTAuth
+    } else {
+      const [projectAccessKey, jwtAuth] = args
+      this.projectAccessKey = projectAccessKey || undefined
+      this.jwtAuth = jwtAuth || undefined
```
VojtechVitek, line 38: this is for backward-compatibility?

can we `console.error()` with a warning to update to the new Opts object?

was this implemented a long time ago? do you expect customers using it?

packages/indexer/src/index.ts
```diff
@@ -4,13 +4,40 @@ import { Indexer as IndexerRpc } from './indexer.gen'
...
   constructor(
     hostname: string,
-    public projectAccessKey?: string,
-    public jwtAuth?: string
+    opts: SequenceIndexerOpts
+  );
+  constructor(
+    hostname: string,
+    projectAccessKey?: string,
+    jwtAuth?: string
```
VojtechVitek, line 23: I think this is fine

I'm just thinking - should we call it SecretKey as in the Builder UI? cc @moskalyk

VojtechVitek (approved):
LGTM

Minor suggestions/questions

packages/indexer/src/index.ts
```diff
@@ -4,13 +4,40 @@ import { Indexer as IndexerRpc } from './indexer.gen'
 
 const fetch = typeof global === 'object' ? global.fetch : window.fetch
 
+export interface SequenceIndexerOpts {
+  JWTAuth?: string
+  ProjectAccessKey?: string
+}
+
 export class SequenceIndexer extends IndexerRpc {
+  public jwtAuth?: string
```
VojtechVitek, line 13: I guess we should call it `apiSecretKey`, as this is called "Builder Admin API Secret Keys" in the Builder UI.

VojtechVitek (commented):
https://github.com/0xsequence/issue-tracker/issues/4425

Let's revisit. Do we have this pattern in any other sequence client?

AmandaG133 (approved):

PR #534, opened 2024-06-06
WaaS: native email auth support

patrislav:


packages/waas/src/auth.ts
```diff
@@ -221,24 +223,71 @@ export class SequenceWaaS {
...
+  async signIn(creds: IdTokenIdentity | string, sessionName: string): Promise<SignInResponse> {
+    const idToken = (typeof creds === 'string' ? creds : creds.idToken)
 
-    // Login on WaaS
-    const decoded = jwtDecode(creds.idToken)
+    const intent = await this.waas.signInWithIdToken(idToken)
+    try {
+      const res = await this.registerSession(intent, sessionName)
+
+      await this.waas.completeSignIn({
```
pkieltyka, line 233: is `completeSignIn` like a callback..? or is it for responding with the challenge..?

packages/waas/src/base.ts
```diff
@@ -1,48 +1,53 @@
...
   IntentDataOpenSession,
   IntentDataSendTransaction,
   IntentDataSignMessage,
   IntentDataValidateSession
 } from './clients/intent.gen'
-import { SubtleCryptoBackend, getDefaultSubtleCryptoBackend } from './subtle-crypto'
-import { SecureStoreBackend, getDefaultSecureStoreBackend } from './secure-store'
+import {getDefaultSubtleCryptoBackend, SubtleCryptoBackend} from './subtle-crypto'
+import {getDefaultSecureStoreBackend, SecureStoreBackend} from './secure-store'
+import {keccak256, toUtf8Bytes} from "ethers/lib/utils";
```
pkieltyka, line 49: lets run `pnpm format` :)

packages/waas/src/auth.ts
patrislav: It's an internal method that updates the state to "authenticated" and stores the result of the `registerSession` call. The flow is actually:

1. `initiateAuth` - commit to the login and receive a challenge (e.g. for email auth the code is sent at this point)
2. `registerSession` - respond to the challenge (e.g. for email auth, send the answer)

This `signIn` method is used for legacy OIDC auth that skips the first part of the flow and moves on to `registerSession` immediately.

packages/waas/src/base.ts
patrislav: thanks, I didn't know about that actually 😅

patrislav:
Superseded by #539

PR #539, opened 2024-06-19
WaaS account federation, commit/reveal OIDC, guest accounts & PlayFab

patrislav:
.


packages/waas/src/auth.ts
```diff
@@ -249,50 +255,78 @@ export class SequenceWaaS {
...
+    } else if ('email' in identity) {
+      return this.initiateEmailAuth(identity.email)
+    } else if ('playFabTitleId' in identity) {
+      return this.initiatePlayFabAuth(identity.playFabTitleId, identity.playFabSessionTicket)
+    }
+
+    throw new Error('invalid identity')
+  }
+
+  private async initiateGuestAuth() {
```
taylanpince, line 274: I think we should standard on `init[METHOD]Auth` for naming. It's nicer to keep it the same across the public method and internals.

packages/waas/src/intents/base.ts
```diff
@@ -7,7 +8,7 @@ import { Session } from '../session'
 export type Intent<T> = Omit<RawIntent, 'data' | 'signatures'> & { data: T }
 export type SignedIntent<T> = Omit<RawIntent, 'data'> & { data: T }
 
-const VERSION = '1.0.0'
+const VERSION = `1 (Web ${PACKAGE_VERSION})`
```
pkieltyka, line 11: what is the "1" in front here for..?

patrislav: it's the intent version, perhaps would be better to make it explicit...

pkieltyka: its fine, we can leave it for now, thanks for clarifying

pkieltyka (approved):

PR #554, opened 2024-07-29
sessions: arweave config reader

attente:


packages/sessions/src/trackers/arweave.ts
```diff
@@ -0,0 +1,569 @@
+import { commons, v2 } from '@0xsequence/core'
+import { migrator } from '@0xsequence/migration'
+import { CachedEIP5719 } from '@0xsequence/replacer'
+import { ethers } from 'ethers'
+import { ConfigTracker, PresignedConfig, PresignedConfigLink } from '../tracker'
+
+const RATE_LIMIT_RETRY_DELAY = 5 * 60 * 1000
+
+// depending on @0xsequence/abi breaks 0xsequence's proxy-transport-channel integration test
```
pkieltyka, line 9: this is a strange quirk with pnpm, we will have to rebuild the lock and review. Patryk and I hit it recently. For now as long as your tests pass, then we are good but we will revisit this in the ethers-v6 branch, but it doesn't make sense to do it before that

packages/sessions/src/trackers/arweave.ts
```diff
@@ -0,0 +1,569 @@
...
+      response = await fetch('https://arweave.net/graphql', {
+        method: 'POST',
+        headers: { 'Content-Type': 'application/json' },
+        body: JSON.stringify({ query }),
+        redirect: 'follow'
+      })
+      if (response.status !== 429) {
+        break
+      }
+      await new Promise(resolve => setTimeout(resolve, RATE_LIMIT_RETRY_DELAY))
```
pkieltyka, line 516: could you add a console.warn() or something if we hit a 429? 

"arweave.net client received rate limit request, waiting for ${RATE_LIMIT_RETRY_DELAY/1000)} seconds and trying again..."

PR #560, opened 2024-08-07
Updated intent.gen without webrpc types, errors exported from authenticator.gen

tolgahan-arikan:
We were exporting error types from intent.gen instead of authenticator.gen, which caused issues when trying to check the type of errors, for example `if (error instanceof WebrpcEndpointError)` was not returning true while error was an instance of WebrpcEndpointError, but from authenticator.gen.


packages/waas/src/auth.ts
```diff
@@ -37,7 +37,15 @@ import {
...
-import { WaasAuthenticator, AnswerIncorrectError, Chain, EmailAlreadyInUseError, Session } from './clients/authenticator.gen'
+import {
+  WaasAuthenticator,
+  AnswerIncorrectError,
+  Chain,
+  EmailAlreadyInUseError,
+  Session,
+  WebrpcEndpointError,
+  WebrpcError
+} from './clients/authenticator.gen'
```
attente, line 48: Is this change necessary? What uses `WebrpcEndpointError` and `WebrpcError` in this file?

tolgahan-arikan: yep you are right, I think I forgot to remove after testing

attente (approved):

PR #563, opened 2024-08-13
HTTP signature check for WaasAuthenticator requests

tolgahan-arikan:


packages/waas/src/auth.ts
```diff
@@ -145,6 +127,34 @@ export function defaultArgsOrFail(
...
+    {
+      alg: 'RS256',
+      e: 'AQAB',
+      kid: '9LkLZyHdNq1N2aeHMlC5jw',
+      kty: 'RSA',
+      n: 'qllUB_ERsOjbKx4SirGow4XDov05lQyhiF7Duo4sPkH9CwMN11OqhLuIqeIXPq0rPNIXGP99A7riXTcpRNk-5ZNL29zs-Xjj3idp7nZQZLIU1CBQErTcbxbwUYp8Q46k7lJXVlMmwoLQvQAgH8BZLuSe-Xk1tye0mDC-bHvmrMfqm2zmuWeDnZercU3Jg2iYwyPrjKWx7YSBSMTXTKPGndws4m3s3XIEpI2alLcLLWsPQk2UjIlux6I7vLwvjM_BgjFhYHqgg1tgZUPn_Xxt4wvhobF8UIacRVmGcuyYBnhRxKnBQhEClGSBVtnFYYBSvRjTgliOwf3DhFoXdnmyPQ',
+      use: 'sig'
+    }
+  ]
+}
```
patrislav, line 143: I feel like there should be a way to possibly override this key or disable it entirely. In the current form, it won't work locally or in any potential new environments we may create. So it'd be good to have an option to pass `null` to disable or a key to override.

packages/waas/src/auth.ts
```diff
@@ -171,10 +184,90 @@ export class SequenceWaaS {
...
+
+          const responseBodyTextUint8Array = new TextEncoder().encode(responseBodyText)
+          const responseBodyTextDigest = await this.cryptoBackend!.digest('SHA-256', responseBodyTextUint8Array)
+          const base64EncodedDigest = btoa(String.fromCharCode(...responseBodyTextDigest))
+
+          if (contentDigestSha !== base64EncodedDigest) {
+            throw new Error('Digest mismatch')
+          }
+
+          const message = `"content-digest": ${contentDigest}\n"@signature-params": ${signatureInput.substring(4)}`
```
patrislav, line 241: Let's add a comment making it clear that "we're removing the first 4 characters to trim the `sig=` prefix"

patrislav (dismissed):

packages/waas/src/auth.ts
```diff
@@ -171,10 +185,91 @@ export class SequenceWaaS {
...
+    if (this.signatureVerificationFailed) {
+      throw new Error('Signature verification failed')
+    }
+
+    if (this.cryptoBackend && this.config.disableHttpSignatureCheck !== true && init?.headers) {
+      const headers: { [key: string]: any } = {}
+
+      headers['Accept-Signature'] = 'sig=();alg="rsa-v1_5-sha256"'
+
+      init!.headers = { ...init!.headers, ...headers }
```
attente, line 203: Was there a reason you didn't just do:

```
init.headers['Accept-Signature'] = 'sig=();alg="rsa-v1_5-sha256"'
```

instead?

packages/waas/src/subtle-crypto.ts
```diff
@@ -71,6 +79,16 @@ export class WindowSubtleCryptoBackend implements SubtleCryptoBackend {
     return window.crypto.subtle.generateKey(algorithm, extractable, keyUsages)
   }
 
+  async importKey(
```
attente, line 82: Don't need `async` here.

packages/waas/src/auth.ts
```diff
@@ -171,10 +185,91 @@ export class SequenceWaaS {
...
+          }
+          const signatureBuffer = Uint8Array.from(atob(sig), c => c.charCodeAt(0))
+
+          const verifyResult = await this.cryptoBackend!.verify(algo, key, signatureBuffer, new TextEncoder().encode(message))
+
+          if (!verifyResult) {
+            throw new Error('Signature verification failed, consequent requests will fail')
+          }
+        } catch (e) {
+          this.signatureVerificationFailed = true
```
attente, line 264: So to understand, if any request fails signature validation, then all future attempts are assumed to fail?

packages/waas/src/auth.ts
```diff
@@ -171,10 +185,91 @@ export class SequenceWaaS {
...
+
+      headers['Accept-Signature'] = 'sig=();alg="rsa-v1_5-sha256"'
+
+      init!.headers = { ...init!.headers, ...headers }
+    }
+
+    const response = fetch(input, init)
+
+    if (this.cryptoBackend && this.config.disableHttpSignatureCheck !== true) {
+      response.then(async r => {
```
attente, line 209: I think you need `return` here, no?

attente (changes requested):

packages/waas/src/auth.ts
tolgahan-arikan: yep, if it doesn't validate consequent requests should not be allowed

packages/waas/src/auth.ts
tolgahan-arikan: so I did it this way to make sure I don't add time overhead and only check stuff to validate and leave response as it, I also do `const clone = r.clone()` because otherwise stream will be closed after first read and it can't be used down the line.

packages/waas/src/auth.ts
tolgahan-arikan: no reason, i'll change 👍

packages/waas/src/auth.ts
attente: But this means that the request will appear valid despite having an invalid signature, no? If the signature is invalid, then we treat all future requests as invalid, but why not the first invalid request in this case?

tolgahan-arikan: shared more details on DM

attente: Screenshot 2024-09-03 at 21 22 21

attente (approved):

PR #567, opened 2024-08-29
Removing BigInt.prototype.toJSON in favor of explicit bigintReplacer and bigintReviver

corbanbrook:


packages/utils/src/bigint.ts
```diff
@@ -94,3 +82,26 @@ export const formatUnits = (value: bigint, decimals: number = 18): string => {
...
+  if (typeof value === 'bigint') {
+    return { $bigint: value.toString() }
+  }
+
+  return value
+}
+
+// JSON.parse will need to convert our serialized bigints back into BigInt
+export const bigintReviver = (key: string, value: any): any => {
+  if (value !== null && typeof value === 'object' && '$bigint' in value && typeof value.$bigint === 'string') {
```
attente, line 97: Don't need `'$bigint' in value` since you're checking `typeof`:

```
if (value && typeof value === 'object' && typeof value.$bigint === 'string') {
```

attente (approved):

PR #571, opened 2024-09-11
WaaS isSignedIn method fix

tolgahan-arikan:


marino39 (dismissed):
LGTM

packages/waas/src/auth.ts
```diff
@@ -353,7 +360,29 @@ export class SequenceWaaS {
...
+    if ((await this.waas.isSignedIn()) === false) {
+      return
+    }
+    // if we can fetch sessions from API, then we are signed in
+    // if not and error is the related session error, then we drop the session
+    try {
+      const sessions = await this.listSessions()
+      if (sessions.length > 0) {
+        return
+      }
```
attente, line 374: If `sessions.length === 0` then this does the same thing.

attente: If this is intentional, you can just replace these four lines with:

```
await this.listSessions()
```

attente (approved):

PR #582, opened 2024-09-27
waas: new contractCall transaction type

pkieltyka:


packages/waas/src/clients/intent.gen.ts
```diff
@@ -196,10 +197,16 @@ export interface IntentDataGetIdToken {
 export interface TransactionRaw {
   type: string
   to: string
-  value: string
+  value?: string
```
pkieltyka, line 200: value is optional

packages/waas/src/intents/transactions.ts
```diff
@@ -59,6 +60,16 @@ export type SendERC1155Args = TransactionFeeArgs & {
   data?: string
 }
 
+export type SendContractCallArgs = TransactionFeeArgs & {
+  chainId: number
+  to: string
+  value?: ethers.BigNumberish
+  abi: string
+  func?: string
```
pkieltyka, line 68: func is optional, as its only needed if you pass in the full abi object, if you pass just a function selector, then that implies the function from the selector

packages/waas/src/intents/transactions.ts
```diff
@@ -59,6 +60,16 @@ export type SendERC1155Args = TransactionFeeArgs & {
...
+  args: any[]
```
pkieltyka, line 69: using `any[]` type here, but maybe it will be confusing to some..? the reason is to support nested encoding.. thoughts? maybe we make this an interface type like: `(string | NestedEncode)[]`. .. but, need a better name then NestedEncode :P

packages/waas/tests/intents.spec.ts
```diff
@@ -127,11 +127,10 @@ describe('Payloads', () => {
             data: '0x223344'
           },
           {
-            type: 'delayedEncode',
+            type: 'contractCall',
```
pkieltyka, line 130: using the name "contractCall" for now, since this is inside of a "sendTransaction" above it, but, usually the "call" is related to read-only operations.

We could call it just "contract" or "contractTx" or "contractSend" .. if anyone has any other ideas, please let us know

pkieltyka:
here is an example of how the data structures would look like at the low level: https://github.com/0xsequence/sequence.js/blob/intents-contract-call-type/packages/waas/tests/intents.spec.ts#L129-L145

```typescript
          {
            type: 'contractCall',
            to: '0x140d72763D1ce39Ad4E2e73EC6e8FC53E5b73B64',
            data: {
              abi: 'fillOrKillOrder(uint256 orderId, uint256 maxCost, address[] fees, bytes data)',
              args: [
                '48774435471364917511246724398022004900255301025912680232738918790354204737320',
                '1000000000000000000',
                '["0x8541D65829f98f7D71A4655cCD7B2bB8494673bF"]',
                {
                  abi: 'notExpired(uint256,string)',
                  args: ['1600000000', 'Nov 1st, 2020']
                }
              ]
            }
          }
```

of coure this `fillOrKillOrder` is intentionally complicated to show off: a.) arguments passed as strings b.) array type support c.) nested encoding

the "args" portion is the same as before, and when I look at this.. I wonder if its correct and if it can be improved..
1. The array argument is specified as `'["0x8541D65829f98f7D71A4655cCD7B2bB8494673bF"]'` which is a string encoded array with string contents.. could we not just make it, `["0x8541D65829f98f7D71A4655cCD7B2bB8494673bF"]`, which means, its not double encoded...?
2. the nested encode of `notExpired(uint256,string)` is cool, but I noticed from in the Go tests, that the encoded result will include the "notExpired" function selector into the encoding, is that intentional? My thought is we'd only want "(uint256,string)" as the tuple for encoding the data into a "bytes" type, is there ever a time we'd want to keep "notExpired" encoded in there as well...? feels like a bug, or I'm forgetting/misunderstanding something, cc: @marino39 as you wrote the original implementation of this in go-sequence/intents – I'm happy to correct it for us if we all decide what we want

Agusx1211:
> the nested encode of notExpired(uint256,string) is cool, but I noticed from in the Go tests, that the encoded result will include the "notExpired" function selector into the encoding

I think including the function selector is correct, if we want the other scenario we could add the option of passing `(uint256,string)` (without a function name) and then encode it without the function selector.

The original use-case for this was calling functions that are triggered from things like ERC-1155 callbacks

packages/waas/src/intents/transactions.ts
patrislav: I will always support stricter typing instead of `any`, so +1 from me to type it properly. I'd suggest having a `ContractCallArgument` (or similar) type that would be `string | ...`

patrislav (commented):
LGTM!

pkieltyka:
hey all, should be ready to go, and go-sequence has been updated to remove the need of double-encoding array types.

see https://github.com/0xsequence/sequence.js/blob/intents-contract-call-type/packages/waas/tests/intents.spec.ts#L129-L145

```
{
  type: 'contractCall',
  to: '0x140d72763D1ce39Ad4E2e73EC6e8FC53E5b73B64',
  data: {
    abi: 'fillOrKillOrder(uint256 orderId, uint256 maxCost, address[] fees, bytes data)',
    args: [
      '48774435471364917511246724398022004900255301025912680232738918790354204737320',
      '1000000000000000000',
      ['0x8541D65829f98f7D71A4655cCD7B2bB8494673bF'],
      {
        abi: 'notExpired(uint256,string)',
        args: ['1600000000', 'Nov 1st, 2020']
      }
    ]
  }
}
```

is what clients will pass

packages/waas/src/intents/transactions.ts
```diff
@@ -59,6 +60,22 @@ export type SendERC1155Args = TransactionFeeArgs & {
...
+export type SendContractCallArgs = TransactionFeeArgs & {
+  chainId: number
+  to: string
+  value?: ethers.BigNumberish
+  abi: string
+  func?: string
+  args: ContractCallArguments[]
+}
+
+export type ContractCallArguments = string | {
```
pkieltyka, line 72: "args" can either just be the argument as a string, or it can be another object which represents nested encoding to happen on server-side

PR #593, opened 2024-10-17
Prevent using same session id to sign in multiple times

tolgahan-arikan:


packages/waas/src/auth.ts
```diff
@@ -383,7 +383,13 @@ export class SequenceWaaS {
     return this.waas.isSignedIn()
   }
 
-  signIn(creds: Identity, sessionName: string): Promise<SignInResponse> {
+  async signIn(creds: Identity, sessionName: string): Promise<SignInResponse> {
+    // We clear and drop session regardless of whether it's signed in or not
+    const currentSessionId = await this.waas.getSessionId()
+    if (currentSessionId) {
+      await this.dropSession({ sessionId: currentSessionId, strict: false })
```
pkieltyka, line 390: does .dropSession make a call to the waas service..? what if this raises an exception / fails..?

pkieltyka: Im just mentioning so we ensure it fails gracefully etc

tolgahan-arikan: `strict: false` allows it to do it gracefully and continue to clear stuff regardless

```ts
  async dropSession({ sessionId, strict }: { sessionId?: string; strict?: boolean } = {}) {
    const thisSessionId = await this.waas.getSessionId()
    if (!thisSessionId) {
      throw new Error('session not open')
    }

    const closeSessionId = sessionId || thisSessionId

    try {
      const intent = await this.waas.signOutSession(closeSessionId)
      const result = await this.sendIntent(intent)

      if (!isCloseSessionResponse(result)) {
        throw new Error(`Invalid response: ${JSON.stringify(result)}`)
      }
    } catch (e) {
      if (strict) {
        throw e
      }

      console.error(e)
    }

    if (closeSessionId === thisSessionId) {
      if (!this.secureStoreBackend) {
        throw new Error('No secure store available')
      }

      const session = await newSessionFromSessionId(thisSessionId, this.cryptoBackend, this.secureStoreBackend)
      session.clear()
      await this.waas.completeSignOut()
      await this.deviceName.set(undefined)
    }
  }
  ```

File: packages/waas/src/auth.ts [hunk 2]
```diff
@@ -363,6 +363,8 @@ export class SequenceWaaS {
   private async updateSessionStatus() {
     // if we are not signed in, then we don't need to check and update session status
     if ((await this.waas.isSignedIn()) === false) {
```
pkieltyka, line 365: does `isSignedIn()` make any remote network requests? or is it all local..?

packages/waas/src/auth.ts
```diff
@@ -363,6 +363,8 @@ export class SequenceWaaS {
...
+      // just to be safe, remove any possible data related to previous session
+      await this.waas.completeSignOut()
```
pkieltyka, line 367: does `completeSignOut()` maybe any network requests or is it all local?

packages/waas/src/auth.ts
```diff
@@ -363,6 +363,8 @@ export class SequenceWaaS {
   private async updateSessionStatus() {
```
pkieltyka, line 363: what calls the `updateSessionStatus()` method?

tolgahan-arikan: isSIgnedIn calls it

packages/waas/src/auth.ts
tolgahan-arikan: all local

pkieltyka (approved):

PR #595, opened 2024-10-23
Add publishWitnessFor

Agusx1211:
Add `publishWitnessFor` that signs witnesses only for a specific set of signers. This should be used when adding new signers, to ensure that they correctly sign a witness.

### DIFF
--- a/packages/account/src/account.ts
+++ b/packages/account/src/account.ts
@@ -490,7 +490,7 @@ export class Account {
     return coder.trim(chainedSignature)
   }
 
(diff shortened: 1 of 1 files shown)


packages/account/src/account.ts
```diff
@@ -490,6 +490,32 @@ export class Account {
     return coder.trim(chainedSignature)
   }
 
+  async publishWitnessFor(signers: string[], chainId: ethers.BigNumberish = 42161): Promise<void> {
```
attente, line 493: Any reason for 42161? What about just 1 or even 0?

attente (approved):

attente (approved):

PR #605, opened 2024-11-06
use /status to get time drift before sending any intents

patrislav:


packages/waas/src/auth.ts
```diff
@@ -533,6 +548,8 @@ export class SequenceWaaS {
   }
 
   private async registerSession(intent: SignedIntent<IntentDataOpenSession>, name: string) {
+    this.updateTimeDrift()
```
pkieltyka, line 551: i think you want `await this.updateTimeDraft()`

packages/waas/src/intents/base.ts
```diff
@@ -11,9 +11,21 @@ export type SignedIntent<T> = Omit<RawIntent, 'data'> & { data: T }
 const INTENTS_VERSION = 1
 const VERSION = `${INTENTS_VERSION} (Web ${PACKAGE_VERSION})`
 
+// The result of localTime - serverTime, in seconds
+let timeDrift: number | undefined
+
+export function getTimeDrift() {
+  return timeDrift
```
pkieltyka, line 18: can you add check for window.sessionStorage (if exists), and try to get from there if it does, otherwise use just the in-memory. sessionStorage is preserved between page reloads, it will just save some unnecessary requests and its a few extra lines to do here and `updateTimeDrift`

packages/waas/src/auth.ts
```diff
@@ -334,7 +336,20 @@ export class SequenceWaaS {
     }
   }
 
+  private async updateTimeDrift() {
+    if (getTimeDrift() === undefined) {
+      const res = await fetch(`${this.config.rpcServer}/status`)
+      const date = res.headers.get('Date')
+      if (!date) {
+        throw new Error('missing Date header value')
```
pkieltyka, line 344: change error to, `failed to get Date header value from /status` ... and probably better is to have a well-defined error in the waas ridl for any error type.. but the above message will do for now

packages/waas/src/auth.ts
```diff
@@ -334,7 +336,20 @@ export class SequenceWaaS {
     }
   }
 
+  private async updateTimeDrift() {
+    if (getTimeDrift() === undefined) {
+      const res = await fetch(`${this.config.rpcServer}/status`)
```
pkieltyka, line 341: use POST, and see other PR comment

pkieltyka (changes requested):

packages/waas/src/intents/base.ts
pkieltyka: like so..

```ts
function isSessionStorageAvailable() {
  return typeof window === 'object' && typeof window.sessionStorage === 'object'
}

const timeDriftKey = '@sequence.timeDrift'

export function getTimeDrift() {
  if (isSessionStorageAvailable()) {
    return window.sessionStorage.get('timeDriftKey') // TODO: convert to number
  } else {
    return timeDrift
  }
}


// etc..
```

packages/waas/src/intents/base.ts
```diff
@@ -11,9 +11,22 @@ export type SignedIntent<T> = Omit<RawIntent, 'data'> & { data: T }
 const INTENTS_VERSION = 1
 const VERSION = `${INTENTS_VERSION} (Web ${PACKAGE_VERSION})`
 
+export function getTimeDrift() {
+  const drift = sessionStorage.getItem('timeDrift')
```
pkieltyka, line 15: your code will throw in react native and node

packages/waas/src/intents/base.ts
```diff
@@ -11,9 +11,34 @@ export type SignedIntent<T> = Omit<RawIntent, 'data'> & { data: T }
...
+let timeDrift: number | undefined
+const timeDriftKey = '@sequence.timeDrift'
+
+function isSessionStorageAvailable() {
+  return typeof window === 'object' && typeof window.sessionStorage === 'object'
+}
+
+export function getTimeDrift() {
+  if (isSessionStorageAvailable()) {
+    const drift = sessionStorage.getItem(timeDriftKey)
```
pkieltyka, line 23: should we use `window.sessionStorage.getItem(..)` to be explicit..? the "window" part that is..?

PR #611, opened 2024-11-25
[OUTDATED] Passkey support

ScreamingHawk:
Major changes:

* Add passkey package with passkey signer
* Use `/rpc/API/GetAuthToken2` with reference chain id for auth
* Use `/rpc/API/SaveSignerSignatures2` with reference chain id for recording witnesses
* Optionally skip EIP-1271 validation when recovering config from signature
* Optionally signer can return a "validation signature" to use for validation. (e.g. Using 6492)
* Guard uses "validation signature" when available for checking validity. See https://github.com/0xsequence/go-sequence/pull/177 

Note: All changes are backwards compatible.


packages/account/src/account.ts
```diff
@@ -665,7 +669,8 @@ export class Account {
     await this.tracker.savePresignedConfiguration({
       wallet: this.address,
       nextConfig: config,
-      signature
+      signature,
+      validateBehavior: 'ignore'
```
Agusx1211, line 673: Why are we using `ignore` here? what's the difference?
I understand next line should caught it anyway, but I still wonder.

packages/core/src/v2/signature.ts
```diff
@@ -251,9 +252,11 @@ export async function recoverTopology(
         throw new Error('Dynamic signature leaf without address')
       }
 
-      const isValid = await isValidSignature(unrecovered.address, subdigest, unrecovered.signature, provider)
-      if (!isValid) {
-        throw new InvalidSignatureLeafError(unrecovered)
+      if (provider) {
```
Agusx1211, line 255: What is the need for adding this? in general, these end up as footguns, as someone using this function never realizes that the signature is never being validate

packages/network/src/json-rpc-provider.ts
```diff
@@ -26,9 +26,9 @@ export interface JsonRpcProviderOptions {
 
 // JsonRpcProvider with a middleware stack. By default it will use a simple caching middleware.
 export class JsonRpcProvider extends ethers.JsonRpcProvider implements EIP1193Provider, JsonRpcSender {
-  #chainId?: number
-  #nextId: number = 1
-  #sender: EIP1193Provider
+  private chainId?: number
```
Agusx1211, line 29: Was this giving any trouble somewhere else? I am not sure if using real private fields was needed in the first place, or just someone playing with new features

packages/sessions/src/trackers/remote/index.ts
```diff
@@ -56,21 +56,36 @@ export class RemoteConfigTracker implements ConfigTracker, migrator.PresignedMig
...
+      }) as string[] | SignerSignature[]
     }
 
-    await this.sessions.saveSignerSignatures({
-      wallet: args.wallet,
-      digest: args.digest,
-      chainID: numberString(args.chainId),
-      signatures: filteredSignatures
-    })
+    if (filteredSignatures.length === 0 || typeof args.signatures[0] === 'string') {
```
Agusx1211, line 73: Why not use `saveSignerSignatures2` every time? This check seems a bit buggy, as it assumes that the whole list of signatures will use the same type.

packages/sessions/src/trackers/remote/index.ts
```diff
@@ -56,21 +56,36 @@ export class RemoteConfigTracker implements ConfigTracker, migrator.PresignedMig
...
   }): Promise<void> {
     let filteredSignatures = args.signatures
     if (this.onlyRecoverable) {
       filteredSignatures = filteredSignatures.filter(signature => {
-        return commons.signer.canRecover(signature)
-      })
+        if (typeof signature === 'string') {
+          return commons.signer.canRecover(signature)
+        } else {
+          // We "recover" using the included address
```
Agusx1211, line 67: What does this mean?

packages/sessions/src/tracker.ts
```diff
@@ -6,6 +6,7 @@ export type PresignedConfig = {
   nextConfig: commons.config.Config
   signature: string
   referenceChainId?: string
+  validateBehavior?: 'ignore' | 'throw'
```
Agusx1211, line 9: Shouldn't this be an argument of `savePresignedConfiguration` instead? Does `PresignedConfigLink` makes any use of it?

packages/signhub/src/signers/signer.ts
```diff
@@ -25,6 +25,11 @@ export interface SapientSigner {
    */
   sign(message: ethers.BytesLike, metadata: object): Promise<ethers.BytesLike>
 
+  /**
+   * Build a validation signature for an undeployed contract signer.
+   */
+  buildValidationSignature?(signatureBytes: ethers.BytesLike): Promise<ethers.BytesLike | undefined>;
```
Agusx1211, line 31: Is this a 6492 signature or something else? if so, can we add 6492 to the comment?

packages/network/src/json-rpc-provider.ts
ScreamingHawk: I was having compile errors. Probably an issue on my end but didn't see the reason to keep it when we don't follow this pattern anywhere else.

packages/sessions/src/trackers/remote/index.ts
ScreamingHawk: Poorly worded comment. I'll remove it. 
The backend only supports `EIP712` and `ETH_SIGN` types for `SaveSignerSignatures` as it uses recovery, but uses `signature.address` for SaveSignerSignatures2` instead.

packages/sessions/src/trackers/remote/index.ts
ScreamingHawk: I wanted to keep `saveWitnesses` backwards compatible. 
I'll update this to check that the entire list matches a format. To make this error happen they would have to ignore the type `string[] | SignerSignature[]` though. We generally accept that the types are as specified.

packages/signhub/src/signers/signer.ts
ScreamingHawk: It is for now but I guess it doesn't have to be. If we extend support to other validation encodings (zk or something?) we could reuse this method. 
Happy to rename if you think we won't reuse later.

packages/core/src/v2/signature.ts
ScreamingHawk: I think it shouldn't be validated by default. `recoverTopology` is about obtaining the tree, not validating the sub signatures. 
Maybe an additional argument would make this clearer.

packages/sessions/src/tracker.ts
ScreamingHawk: Seems related to `referenceChainId`. Both are for validation purposes.

packages/network/src/json-rpc-provider.ts
Agusx1211: This change was made by @corbanbrook. Let’s have him chime in, just in case there was a specific need for it.

packages/sessions/src/trackers/remote/index.ts
Agusx1211: what does it mean backwards compatible? like being able to roll this sequence.js update before sessions is updated?

packages/sessions/src/tracker.ts
Agusx1211: Not really. referenceChainId specifies the chain to use for signature validation (without it, validation wouldn’t be possible), as it gives meaning to the signature. validateBehavior is not directly tied to the signature itself but acts as a rule for the process involving the signature. This “rule” shouldn’t inadvertently cross into other contexts.

packages/signhub/src/signers/signer.ts
Agusx1211: Let’s rename it for now. I’d prefer it to be obvious at a glance.

packages/network/src/json-rpc-provider.ts
Agusx1211: I spoke with Corban, and he doesn’t recall any specific reason for using the private fields. It’s okay to proceed with reverting the change to using `private`.

packages/core/src/v2/signature.ts
ScreamingHawk: Fair call. The difference with EIP-1271 validation is that it may be conditional on factors outside the signature but ECDSA is not. The function assumes the current point in time is the correct baseline for validation, which it may not be (and isn't for this case). 

I don't think we want to be passing transactions around for this so how about changing the signature to:
```ts
function recoverTopology(
  unrecovered: UnrecoveredTopology,
  subdigest: string,
  provider: ethers.Provider, // Require this so it's more likely to be used
  performERC1271Validation = true // Add this. Default true
)
```

packages/account/src/account.ts
ScreamingHawk: This is to prevent the issue with validating EIP-1271 signatures mentioned in the other comments.

packages/core/src/v2/signature.ts
Agusx1211: That works for me

packages/sessions/src/trackers/remote/index.ts
Agusx1211: But couldn't you recover the address if `string[]` and then use `saveSignerSignatures2`?

ScreamingHawk: I don't think so? Correct me if I'm wrong but we'd need the wallet config version, which this function doesn't have access to. We could assume, but that seems even more error prone than using the current method as is.

packages/sessions/src/trackers/local.ts
```diff
@@ -235,7 +235,12 @@ export class LocalConfigTracker implements ConfigTracker, migrator.PresignedMigr
     const savePayload = this.savePayload({ payload })
     const saveNextConfig = this.saveWalletConfig({ config: args.nextConfig })
 
-    const recovered = await v2.signature.SignatureCoder.recover(decoded, payload, this.provider)
+    const recovered = await v2.signature.SignatureCoder.recover(
+      decoded,
+      payload,
+      this.provider,
+      'ignore'
```
Agusx1211, line 242: Why are we passing "ignore" here if we do have a provider? aren't we risking contaminating the local database of signatures?

I guess the alternative would be to miss data, which may be worse.

Agusx1211:
Lets test this on `fun` before merging it to master. I published a tag release `0.0.0-20241216114019`.

Arumi007:
git pull https://github.com/0xsequence/sequence.js/pull/611#discussion_r1886658696

ScreamingHawk:
Converting to draft for reprioritization

pkieltyka:
NOTE: this PR is no longer relevant, as the work has shifted to v3

ScreamingHawk:
Closing. Replaced in v3

PR #614, opened 2024-11-29
waas: use getLocalTime instead of Date.now()

patrislav:


packages/waas/src/intents/base.ts
```diff
@@ -18,6 +18,10 @@ function isSessionStorageAvailable() {
   return typeof window === 'object' && typeof window.sessionStorage === 'object'
 }
 
+export function getLocalTime() {
+  return new Date().getTime()
```
pkieltyka, line 22: interesting.. `new Date().getTime()` works differently then `Date.now()` ?

pkieltyka: I've confirmed, `Date.now()` is very weird, and does not consider day light savings

pkieltyka (approved):

PR #621, opened 2024-12-06
Rename `chainID` into `chainId`

xiam:
This PR will rename `chainID` properties into `chainId`, which is the preferred spelling.


pkieltyka:
we cannot merge this, as it will break too many apps.. but we should figure out some day how we can do it

pkieltyka (changes requested):
this will break too many apps

PR #630, opened 2024-12-16
waas: Add support for signTypedData intent

patrislav:


pkieltyka (commented):
Looks good.

Can you also update "waas-ethers" the signTypedData method..? I added to waas-ethers last week the new "signedTypedData" but my implementation is actually a bit wrong, and we should be having it call the new waas package signTypedData as you've newly added in this PR

patrislav:
> Can you also update "waas-ethers" the signTypedData method..? I added to waas-ethers last week the new "signedTypedData" but my implementation is actually a bit wrong, and we should be having it call the new waas package signTypedData as you've newly added in this PR

I did, but it's going to be invalid due to missing `primaryType`, not sure how can I include it as the function signature (defined by the AbstractSigner interface) doesn't provide it as argument... Or is there another way that I'm missing?

pkieltyka:
I did a bit of testing with ethers ..

```typescript
async function main() {
  const typedDataJson = `{
    "types": {
      "ExampleMessage": [
        { "name": "message", "type": "string" },
        { "name": "value", "type": "uint256" },
        { "name": "from", "type": "address" },
        { "name": "to", "type": "address" }
      ]
    },
    "domain": {
      "name": "EIP712Example",
      "version": "1",
      "chainId": 5,
      "verifyingContract": "0xc0ffee254729296a45a3885639AC7E10F9d54979",
      "salt": "0x70736575646f2d72616e646f6d2076616c756500000000000000000000000000"
    },
    "message": {
      "message": "Test message",
      "value": 10000,
      "from": "0xc0ffee254729296a45a3885639AC7E10F9d54979",
      "to": "0xc0ffee254729296a45a3885639AC7E10F9d54979"
    }
  }`

  const typedData = JSON.parse(typedDataJson)

  const wallet = ethers.Wallet.fromPhrase(mnemonic)
  console.log('wallet address:', wallet.address)

  const domainHashStruct = ethers.TypedDataEncoder.hashDomain(typedData.domain)
  console.log('domainHashStruct:', domainHashStruct)

  const signature = await wallet.signTypedData(typedData.domain, typedData.types, typedData.message)
  console.log('signature:', signature)

  // verify signature
  const recoveredAddress = ethers.verifyTypedData(typedData.domain, typedData.types, typedData.message, signature)
  console.log('recovered address:', recoveredAddress)
}

main()
```

this works fine.

as you can see "primaryType" is not defined. It works because ethers will set the "primaryType" automatically as there is just a single type of "ExampleMessage".

If you add another type under "types" (ie. Person), and do not specify "primaryType" then ethers will respond with the error of 'ambiguous primary types or unused types: "ExampleMessage", "Person"'

I'll update ethkit so we do the same thing.

HOWEVER.. this is still super confusing because surely ethers can accept a "primaryType" somehow. I'll check ethers source to see how it handles the primaryType to be passed and I'll update the PR above myself.

pkieltyka:
this will do it: https://github.com/0xsequence/ethkit/pull/151

PR #631, opened 2024-12-17
Updating initWallet to use connected network configs if they exist

corbanbrook:
Previously dapp side would just use the default network config which pointed to production rpcUrls, etc. So there was a disconnect between the what the specified wallet was using and the dapp, so a dev project access key that would work on wallet side would not work on the dapp side production services. 

This corrects it by making sure to merge in the connected sessions network config when returning providers

### DIFF
--- a/packages/provider/src/init.ts
+++ b/packages/provider/src/init.ts
@@ -84,58 +84,31 @@ export const initWallet = (projectAccessKey: string, partialConfig?: Partial<Pro
     }
   }
 
-  const rpcProviders: Record<number, ethers.JsonRpcProvider> = {}
+  let networks: NetworkConfig[] = []
 
-  // Find any new networks that aren't already defined in sequence.js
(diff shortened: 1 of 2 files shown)


Agusx1211 (approved):
LGTM

packages/provider/src/init.ts
```diff
@@ -84,58 +84,31 @@ export const initWallet = (projectAccessKey: string, partialConfig?: Partial<Pro
...
+
+    // Append projectAccessKey to network rpcUrls
+    networks = networks.map(network => {
+      // Don't double-append in the case the user has already included their access key in the rpc URL
       if (network.rpcUrl.includes(projectAccessKey)) {
         return network
       }
 
-      // this will probably break non-sequence RPC provider URLs.
+      // XXX: This will probably break non-sequence RPC provider URLs.
```
pkieltyka, line 99: lets just search for "sequence.app" in the url then, and skip adding this if thats the case..?

PR #632, opened 2024-12-19
Add immutable wallet proxy hook

ScreamingHawk:
Requires: https://github.com/0xsequence/wallet-contracts/pull/191

Immutable ERC721/1155s have an [OperatorAllowlist](https://github.com/immutable/contracts/blob/main/contracts/allowlist/OperatorAllowlistUpgradeable.sol#L153) that prevent contracts owning tokens. For wallets to own tokens they must have an approved bytecode and implementation address returned by a `PROXY_getImplementation()` function. 

This PR adds the `PROXY_getImplementation()` function to a wallet via the hooks module. This is added whenever the user interacts with an immutable chain. The hook is only added once.

### DIFF
--- a/packages/account/src/account.ts
+++ b/packages/account/src/account.ts
@@ -413,7 +413,8 @@ export class Account {
     }
 
     // On immutable chains, we add the WalletProxyHook
-    if (chainId === ChainId.IMMUTABLE_ZKEVM || chainId === ChainId.IMMUTABLE_ZKEVM_TESTNET) {
+    const { proxyImplementationHook } = this.contexts[status.config.version]
+    if (proxyImplementationHook && (chainId === ChainId.IMMUTABLE_ZKEVM || chainId === ChainId.IMMUTABLE_ZKEVM_TESTNET)) {
       const provider = this.providerFor(chainId)
       if (provider) {
         const hook = new ethers.Contract(this.address, walletContracts.walletProxyHook.abi, provider)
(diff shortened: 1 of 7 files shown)


packages/account/src/account.ts
```diff
@@ -402,13 +402,45 @@ export class Account {
...
+          const hooksInterface = new ethers.Interface(walletContracts.moduleHooks.abi)
+          const tx: commons.transaction.Transaction = {
+            to: this.address,
+            data: hooksInterface.encodeFunctionData(hooksInterface.getFunction('addHook')!, [
+              '0x90611127',
+              '0x1f56dbAD5e8319F0DE9a323E24A31b5077dEB1a4'
+            ]),
+            gasLimit: 50000, // Expected ~28k gas. Buffer added
+            delegateCall: false,
+            revertOnError: false,
```
ScreamingHawk, line 438: `revertOnError: false,` is intentional. We don't want a failure adding the hook to prevent whatever the user's intended transaction is.

packages/account/src/account.ts
```diff
@@ -402,13 +402,45 @@ export class Account {
...
+          console.log('Error getting implementation address', e)
+        }
+        if (!implementation || implementation === ethers.ZeroAddress) {
+          console.log('Adding wallet proxy hook')
+          const hooksInterface = new ethers.Interface(walletContracts.moduleHooks.abi)
+          const tx: commons.transaction.Transaction = {
+            to: this.address,
+            data: hooksInterface.encodeFunctionData(hooksInterface.getFunction('addHook')!, [
+              '0x90611127',
+              '0x1f56dbAD5e8319F0DE9a323E24A31b5077dEB1a4'
```
Agusx1211, line 434: Lets move this hardcoded address somewhere else, maybe context.

ScreamingHawk: Updated

ScreamingHawk: Updated. Now requires this change to wallet webapp to pass context: https://github.com/0xsequence/wallet-webapp/pull/807

Agusx1211:
LGTM Now

PR #689, opened 2025-03-05
feat: add auto prefixing check to account.signMessage

VGabriel45:


tolgahan-arikan (approved):
lgtm! linked locally to webapp and tested there as well and did not see any issues

PR #700, opened 2025-03-22
Create networkNames.md

bunnybones1:
Since we're improving boilerplate configuration, we recommend users use network name slugs in they .env configs.
There's no simple reference to look up those names right now. We could maybe find a better home for this information, but this is a good place to start IMHO

### DIFF
--- a/packages/network/networkNames.md
+++ b/packages/network/networkNames.md
@@ -1,0 +1,61 @@
+# Network Names
+
+Though networks have proper full names, we sometimes need to refer to a chain by a name slug.
+
+A name slug is the name, sometimes simplified, and in kebab-case. This maximizes compatibility in the event that you want to use chain names as url paths, or filenames.
+
+You can use this file as a reference to look up the correct network name slug a network's full name or its chain ID.
+
+| Network Name | Chain ID | Name Slug |
+| --- | --- | --- |
+| Ethereum | 1 | mainnet |
+| Optimism | 10 | optimism |
+| Telos | 40 | telos |
+| Telos Testnet | 41 | telos-testnet |
+| BNB Smart Chain | 56 | bsc |
+| BNB Smart Chain Testnet | 97 | bsc-testnet |
+| Gnosis Chain | 100 | gnosis |
+| Polygon | 137 | polygon |
+| XR1 | 273 | xr1 |
+| Polygon zkEVM | 1101 | polygon-zkevm |
+| Moonbeam | 1284 | moonbeam |
+| Moonbase Alpha | 1287 | moonbase-alpha |
+| Soneium | 1868 | soneium |
+| Soneium Minato (Testnet) | 1946 | soneium-minato |
+| B3 Sepolia | 1993 | b3-sepolia |
+| XR Sepolia | 2730 | xr-sepolia |
+| LAOS | 6283 | laos |
+| The Root Network | 7668 | rootnet |
+| The Root Network Porcini Testnet | 7672 | rootnet-porcini |
+| B3 | 8333 | b3 |
+| Base (Coinbase) | 8453 | base |
+| MONAD Testnet | 10143 | monad-testnet |
+| Immutable zkEVM | 13371 | immutable-zkevm |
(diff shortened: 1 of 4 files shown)


scripts/update-network-names-doc.ts
```diff
@@ -0,0 +1,37 @@
...
+}
+
+lines.push(""); 
+
+networkObjs.sort((a, b) => a.title?.localeCompare(b.title || "") || 0);
+lines.push(`| Chain Name | Name Slug |`);
+lines.push(`| --- | --- |`);
+for (const n of networkObjs) {
+  lines.push(`| ${n.title} | ${n.name} |`);
+}
```
VojtechVitek, line 31: Can we combine `id | slug | name` into a single table?

packages/network/networkNames.md
```diff
@@ -0,0 +1,133 @@
...
+A name slug is the name, sometimes simplified, and in kebab-case. This maximizes compatibility in the event that you want to use chain names as url paths, or filenames.
+
+You can use this file as a reference to look up the correct network name slug a network's full name or its chain ID.
+
+| Chain ID | Name Slug |
+| --- | --- |
+| 1 | mainnet |
+| 3 | ropsten |
+| 4 | rinkeby |
+| 5 | goerli |
```
VojtechVitek, line 14: Some of these are obsolete / unsupported chains.

I think we should mark them as such.

VojtechVitek (commented):
Good idea! :) Can we reference this auto-generated file from top-level README too?

scripts/update-network-names-doc.ts
bunnybones1: Yeah, I was waffling about that because chain ids and alphabetical name are different orders, and I wasn't sure which one users would prefer, so I put both.
But that said, users can use CMD+F honestly.
How about 
`Name | ID | slug`
So that the name, alphabetically, will be the primary sort

packages/network/networkNames.md
bunnybones1: Yeah, now that this doc is autogenerated from the pacakge/network/constants, it should either be removed there or marked somehow as deprecated so we can omit it from this new doc. Or we could leave it for now, not a rush

scripts/update-network-names-doc.ts
VojtechVitek: I'll leave order up to you, whatever makes more sense from user perspective :)

taylanpince: Let's combine into a single table please @bunnybones1 it will be easier to parse

packages/network/networkNames.md
taylanpince: @bunnybones1 They are already marked as deprecated, see here: https://github.com/0xsequence/devops/blob/b0f519669e4ec25963125c2ba0618d8d1c5a360f/constants.ts#L222

Can you have the script parse this flag and ignore if it's `deprecated: true`

VojtechVitek (approved):
LGTM

packages/network/networkNames.md
```diff
@@ -0,0 +1,70 @@
...
+A name slug is the name, sometimes simplified, and in kebab-case. This maximizes compatibility in the event that you want to use chain names as url paths, or filenames.
+
+You can use this file as a reference to look up the correct network name slug a network's full name or its chain ID.
+
+| Network Name | Chain ID | Name Slug |
+| --- | --- | --- |
+| Ethereum | 1 | mainnet |
+| Ropsten | 3 | ropsten |
+| Rinkeby | 4 | rinkeby |
+| Goerli | 5 | goerli |
```
VojtechVitek, line 14: This one is deprecated. Please filter out deprecated networks.. or mark them as ~deprecated~.

packages/network/networkNames.md
bunnybones1: done!

packages/network/networkNames.md
bunnybones1: done!

taylanpince (approved):
Beautiful, thank you

PR #760, opened 2025-05-22
Change demo to donate to give directly

Agusx1211:


shunkakinoki (approved):
Maybe split between `0xe1b15a827584b635f57c046592a4a021abad628c` and `peterk.eth` too?

PR #794, opened 2025-06-12
Remove anypay

taylanpince:


miguelmota:
looked good, thanks @taylanpince!

PR #798, opened 2025-06-19
Permissions Builder

ScreamingHawk:
Adds a helper class for building permissions without having to worry about the bytes encoding. 

```ts
const permission = PermissionBuilder.for(usdc_address)
  .forFunction('transfer(address to, uint256 value)')
  .withAddressParam('to', recipient_address)
  .build()
```

```ts
ERC20PermissionBuilder.buildApprove(erc20Address, spenderAddress, approveAmount)
```

More example usage in the tests. 

- [X] Support unrestricted permissions
- [X] Support primitive types
- [X] Support one time use permissions
- [x] Support dynamic types
- [x] More helpers for specific ERC20/721/1155 functions


ScreamingHawk:
Note this feature does not support functions with multiple dynamic types. This is because the encoding of the calldata is impossible to predict without knowing all values that will be sent in the call. 

We _could_ support this if we also perform transaction encoding (or reencoding) in the SDK as well. Out of scope for this ticket.

PR #799, opened 2025-06-19
initial guard v3 integration

patrislav:


packages/wallet/wdk/src/sequence/guard.ts
```diff
@@ -0,0 +1,72 @@
...
+  async sign(
+    wallet: Address.Address,
+    chainId: bigint,
+    payload: Payload.Payload,
+  ): Promise<SequenceSignature.SignatureOfSignerLeafHash> {
+    const digest = Payload.hash(wallet, chainId, payload)
+    const typedData = Payload.toTyped(wallet, chainId, payload)
+    const serialized = Hex.fromString(TypedData.serialize(typedData))
+
+    const auxData = AbiParameters.encode(AbiParameters.from(['address', 'uint256', 'bytes', 'bytes']), [
```
Agusx1211, line 18: It may not be the best time now but it would be nice to get rid of the concept of `auxData`, and instead just pass these arguments on the RPC request directly. It is something that no longer makes sense.

packages/wallet/wdk/src/sequence/wallets.ts
```diff
@@ -525,6 +526,11 @@ export class Wallets {
     // Sign witness using the passkey signer
     await loginSigner.signer.witness(this.shared.sequence.stateProvider, wallet.address, loginSigner.extra)
 
+    // Sign witness using the guard signer
+    if (guardTopology) {
+      await this.shared.modules.guard.witness(wallet.address)
```
Agusx1211, line 531: I figure the only reason we are signing a witness with the guard is to be able to "know" what kind of signer is the guard signer later, correct?

patrislav: Tbh, I don't know, I thought that every signer must sign a witness, is that not the case? Is it only used for figuring out signer kind? In that case, maybe we don't need it, guard address is static so we'll always know which signer is the guard. Though we can do it "just in case"?

Signing (anything) at this point could be useful to ensure that guard is working and accessible. But I don't know if we can do anything about that anyway.

Agusx1211: Lets remove this then, and check that everything still works. Lets not do remote calls and sign operations if we don't know what purpose they serve.

Agusx1211:
Once we solve the witness thing this looks good to me!

PR #802, opened 2025-06-25
Add exact calldata match permission generator

ScreamingHawk:


ScreamingHawk:
We should consider separating the selector from the calldata so we can have a defined selector rule, which would be useful for decoding the permission on the wallet app.

PR #840, opened 2025-08-13
Networks update

ScreamingHawk:
* Networks leverages format used by devops, except:
  * `chain_id` is bigint
  * `rpc` added
  * `NetworkType.LOCAL` added
  * Deprecated `testnet` field removed (using type instead)
* Consistent usage of bigint for ChainId in dapp-client
* Removes 0xsequence/network

Note: Keeps deprecated chains and ids as these were available in devops. Should we remove these?

### DIFF
--- a/packages/wallet/dapp-client/src/utils/index.ts
+++ b/packages/wallet/dapp-client/src/utils/index.ts
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
-import { allNetworks } from '@0xsequence/network'
+import { Network } from '@0xsequence/wallet-primitives'
 import { Bytes, Hex } from 'ox'
 import { NODES_URL, RELAYER_URL } from './constants.js'
 
@@ -117,8 +117,8 @@ function applyTemplate(template: string, values: Record<string, string>) {
   })
 }
 
-export const getNetwork = (chainId: number) => {
-  const network = allNetworks.find((network) => network.chainId === chainId)
+export const getNetwork = (chainId: Network.ChainId | bigint | number) => {
+  const network = Network.getNetworkFromChainId(chainId)
 
   if (!network) {
     throw new Error(`Network with chainId ${chainId} not found`)
@@ -127,7 +127,7 @@ export const getNetwork = (chainId: number) => {
   return network
 }
 
-export const getRpcUrl = (chainId: number) => {
+export const getRpcUrl = (chainId: Network.ChainId | bigint | number) => {
   const network = getNetwork(chainId)
 
   const url = applyTemplate(NODES_URL, { network: network.name })
@@ -135,7 +135,7 @@ export const getRpcUrl = (chainId: number) => {
   return url
 }
 
-export const getRelayerUrl = (chainId: number) => {
+export const getRelayerUrl = (chainId: Network.ChainId | bigint | number) => {
   const network = getNetwork(chainId)
 
   const url = applyTemplate(RELAYER_URL, { network: network.name })
@@ -143,7 +143,7 @@ export const getRelayerUrl = (chainId: number) => {
   return url
 }
 
-export const getExplorerUrl = (chainId: number, txHash: string) => {
+export const getExplorerUrl = (chainId: Network.ChainId | bigint | number, txHash: string) => {
   const network = getNetwork(chainId)
   const explorerUrl = network.blockExplorer?.rootUrl
   if (!explorerUrl) {
--- a/packages/wallet/dapp-client/src/utils/index.ts
+++ b/packages/wallet/dapp-client/src/utils/index.ts
@@ -145,7 +145,7 @@ export const getRelayerUrl = (chainId: Network.ChainId | bigint | number) => {
 
 export const getExplorerUrl = (chainId: Network.ChainId | bigint | number, txHash: string) => {
   const network = getNetwork(chainId)
-  const explorerUrl = network.blockExplorer?.rootUrl
+  const explorerUrl = network.blockExplorer?.url
   if (!explorerUrl) {
     throw new Error(`Explorer URL not found for chainId ${chainId}`)
   }
--- a/packages/wallet/primitives/src/network.ts
+++ b/packages/wallet/primitives/src/network.ts
@@ -16,7 +16,7 @@ export interface Network {
   rpcUrl: string
   logoUrl?: string
   blockExplorer?: BlockExplorerConfig
-  nativeToken: {
+  nativeCurrency: {
     symbol: string
     name: string
     decimals: number
@@ -157,7 +157,7 @@ export const ALL: Network[] = [
       name: 'Etherscan',
       url: 'https://etherscan.io/',
     },
-    nativeToken: {
+    nativeCurrency: {
       symbol: 'ETH',
       name: 'Ether',
       decimals: 18,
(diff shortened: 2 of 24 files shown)


packages/wallet/dapp-client/src/utils/index.ts
```diff
@@ -117,8 +117,8 @@ function applyTemplate(template: string, values: Record<string, string>) {
   })
 }
 
-export const getNetwork = (chainId: number) => {
-  const network = allNetworks.find((network) => network.chainId === chainId)
+export const getNetwork = (chainId: Network.ChainId | bigint | number) => {
```
corbanbrook, line 120: might also want to accept Hex | string here too if this is a helper function that could get a network from any type

packages/wallet/dapp-client/src/utils/index.ts
```diff
@@ -127,23 +127,23 @@ export const getNetwork = (chainId: number) => {
   return network
 }
 
-export const getRpcUrl = (chainId: number) => {
+export const getRpcUrl = (chainId: Network.ChainId | bigint | number) => {
```
corbanbrook, line 130: Not sure if Network.ChainId does anything here since bigint is a broader type. Does this help in intellisense?

packages/wallet/dapp-client/src/utils/index.ts
```diff
@@ -127,23 +127,23 @@ export const getNetwork = (chainId: number) => {
...
   const network = getNetwork(chainId)
 
   const url = applyTemplate(NODES_URL, { network: network.name })
 
   return url
 }
 
-export const getRelayerUrl = (chainId: number) => {
+export const getRelayerUrl = (chainId: Network.ChainId | bigint | number) => {
```
corbanbrook, line 138: Not sure if Network.ChainId does anything here since bigint is a broader type. Does this help in intellisense?

packages/wallet/dapp-client/src/utils/index.ts
```diff
@@ -127,23 +127,23 @@ export const getNetwork = (chainId: number) => {
...
   const network = getNetwork(chainId)
 
   const url = applyTemplate(RELAYER_URL, { network: network.name })
 
   return url
 }
 
-export const getExplorerUrl = (chainId: number, txHash: string) => {
+export const getExplorerUrl = (chainId: Network.ChainId | bigint | number, txHash: string) => {
```
corbanbrook, line 146: Not sure if Network.ChainId does anything here since bigint is a broader type. Does this help in intellisense?

packages/wallet/primitives/src/network.ts
```diff
@@ -1,630 +1,1229 @@
...
+      rootUrl: 'https://porcini.rootscan.io/',
+    },
+    nativeToken: {
+      symbol: 'XRP',
+      name: 'XRP',
+      decimals: 18,
+    },
+  },
+  {
+    chainId: ChainId.HARDHAT,
```
corbanbrook, line 985: We don't need to include these HARDHAT networks in the primitives, if required they can be defined within the test suite only.

packages/wallet/primitives/src/network.ts
```diff
@@ -1,630 +1,1229 @@
...
+    title: 'Hardhat (local testnet)',
+    rpc: 'http://localhost:8545',
+    nativeToken: {
+      symbol: 'ETH',
+      name: 'Ether',
+      decimals: 18,
+    },
+  },
+  {
+    chainId: ChainId.HARDHAT_2,
```
corbanbrook, line 997: We don't need to include these HARDHAT networks in the primitives, if required they can be defined within the test suite only.

packages/wallet/primitives/src/network.ts
```diff
@@ -1,630 +1,1229 @@
-export type Network = {
+export enum NetworkType {
+  MAINNET = 'mainnet',
+  TESTNET = 'testnet',
+  LOCAL = 'local',
```
corbanbrook, line 4: We don't need this LOCAL network type. We should remove the HARDHAT networks from this list as they are for testing only and don't want to be exposed

packages/wallet/primitives/src/network.ts
```diff
@@ -1,630 +1,1229 @@
...
+}
+
+export type BlockExplorerConfig = {
+  name?: string
+  rootUrl: string
```
corbanbrook, line 9: should rename `rootUrl` to  `url` for simplicity. This will then also maps directly to the viem/wagmi BlockExplorer interface

packages/wallet/primitives/src/network.ts
```diff
@@ -1,630 +1,1229 @@
...
   name: string
+  title?: string
   rpc: string
-  chainId: bigint
-  explorer: string
-  nativeCurrency: {
-    name: string
+  logoURI?: string
+  blockExplorer?: BlockExplorerConfig
+  nativeToken: {
```
corbanbrook, line 20: viem/wagmi chain interface calls it `nativeCurrency` rather than `nativeToken` we might want to have consistency

packages/wallet/primitives/src/network.ts
```diff
@@ -1,630 +1,1229 @@
...
+  name?: string
+  rootUrl: string
+}
+
+export interface Network {
+  chainId: ChainId
+  type: NetworkType
   name: string
+  title?: string
   rpc: string
```
corbanbrook, line 17: `rpc` -> `rpcUrl`

packages/wallet/primitives/src/network.ts
```diff
@@ -1,630 +1,1229 @@
...
-  chainId: bigint
-  explorer: string
-  nativeCurrency: {
-    name: string
+  logoURI?: string
```
corbanbrook, line 18: `logoURI` to `logoUrl`

packages/wallet/primitives/src/network.ts
```diff
@@ -1,630 +1,1229 @@
...
-export const SkaleNebula: Network = {
-  name: 'skale-nebula',
-  rpc: getRpcUrl('skale-nebula'),
-  chainId: 1482601649n,
-  explorer: 'https://green-giddy-denebola.explorer.mainnet.skalenodes.com/',
-  nativeCurrency: {
-    name: 'SKALE Fuel',
-    symbol: 'sFUEL',
-    decimals: 18,
+export const ALL: Network[] = [
```
corbanbrook, line 153: Do we uppercase `ALL` here?

packages/wallet/primitives/src/network.ts
```diff
@@ -1,630 +1,1229 @@
...
+  {
+    chainId: ChainId.MAINNET,
+    type: NetworkType.MAINNET,
+    name: 'mainnet',
+    title: 'Ethereum',
+    rpc: getRpcUrl('mainnet'),
+    logoURI: `https://assets.sequence.info/images/networks/medium/${ChainId.MAINNET}.webp`,
```
corbanbrook, line 160: since all logos will use the same url maybe we should create a `getLogoUrl(chainId: ChainId) => ``https://assets.sequence.info/images/networks/medium/${chainId}.webp`` function

`logoUrl: getLogoUrl(ChainId.MAINNET)`

packages/wallet/primitives/src/network.ts
```diff
@@ -1,630 +1,1229 @@
...
-  LaosSigmaTestnet,
-  Moonbeam,
-  MoonbaseAlpha,
-  Etherlink,
-  EtherlinkTestnet,
-  MonadTestnet,
-  Somnia,
-  SomniaTestnet,
-]
+export function getNetworkFromChainId(chainId: ChainId | bigint | number): Network | undefined {
```
corbanbrook, line 1227: Should we also support Hex | string here since we could be getting  from a EIP1193 response?

ScreamingHawk:
I'm on board with updating chain id to number. Bigint has other issues like it requires non standard de/serialization and can't be used as a map key. Can go in another PR

PR #845, opened 2025-08-21
Use guard for sessions

patrislav:
- move guard logic into `@0xsequence/guard` (where it should've been)
- change wallet configuration to wrap each module in a "guard-gated" nested signer
- provide a `signEnvelope` method on SessionManager that adds a guard signature if needed


packages/services/guard/src/index.ts
```diff
@@ -1,2 +1,10 @@
-// export { Guard } from './guard.gen'
-// export * from './signer'
+import { Address, Hex, Bytes, Signature } from 'ox'
+
+export interface GuardSigner {
+  readonly address: Address.Address
+
+  sign(wallet: Address.Address, chainId: number, digest: Bytes.Bytes, message: Hex.Hex): Promise<Signature.Signature>
```
ScreamingHawk, line 6: Can `digest` be derived from the `message`? Do we need both here?

packages/wallet/core/src/signers/session-manager.ts
```diff
@@ -283,6 +285,48 @@ export class SessionManager implements SapientSigner {
...
+          signature: {
+            type: 'hash',
+            ...guardSignature,
+          },
+        })
+      } catch (error) {
+        // We don't throw here because we may have reached threshold with the sapient signature alone
+        console.error('failed to sign with guard', error)
+      }
+    }
```
ScreamingHawk, line 326: I don't think this logic belongs in the session manager. Can we have a guard signer in core that implements these functions instead?
When signing the envelope in the dapp-client, the package can sign with the session manager and then if the signature doesn't met the wallet threshold, pass it to the guard to sign.

packages/wallet/wdk/src/sequence/manager.ts
```diff
@@ -403,7 +403,9 @@ export class Manager {
       cron: new Cron(shared),
       logger: new Logger(shared),
       devices: new Devices(shared),
-      guard: new Guard(shared),
+      guard: shared.sequence.guardUrl
+        ? new Guard.Sequence.GuardSigner(shared.sequence.guardUrl, shared.sequence.guardAddress)
+        : new Guard.Local.GuardSigner(Secp256k1.randomPrivateKey()),
```
ScreamingHawk, line 408: Can we support a fixed local key? Maybe as another env var. Otherwise the guard will be different every time you restart the app making local dev harder.

packages/wallet/wdk/src/sequence/wallets.ts
```diff
@@ -461,6 +482,9 @@ function fromModulesTopology(topology: Config.Topology): Config.SapientSignerLea
     if (topology.address !== Constants.ZeroAddress) {
       throw new Error('signer-leaf-not-allowed-in-modules-topology')
     }
+  } else if (Config.isNestedLeaf(topology) && Config.isNode(topology.tree)) {
+    // This module is wrapped with a guard, it's always going to be the first element
+    modules = [...modules, ...fromModulesTopology(topology.tree[0])]
```
ScreamingHawk, line 487: Can we recursively call `fromModulesTopology` on a nestedLeaf instead? That way we cater to more nested arranged than just this guard set up

packages/wallet/wdk/src/sequence/wallets.ts
```diff
@@ -1097,12 +1121,10 @@ export class Wallets implements WalletsInterface {
...
       ]),
-      guard: guardSigners
-        ? await this.shared.modules.signers.resolveKinds(wallet, [
-            ...guardSigners.signers,
-            ...guardSigners.sapientSigners,
-          ])
-        : [],
+      guard:
+        guardSigners && guardSigners.signers.length > 0
+          ? (await this.shared.modules.signers.resolveKinds(wallet, guardSigners.signers))[0]
```
ScreamingHawk, line 1126: What if multiple different guards? Do we need extra validation here?

ScreamingHawk (changes requested):

packages/wallet/wdk/src/sequence/wallets.ts
```diff
@@ -446,6 +447,26 @@ function toModulesTopology(modules: Config.SapientSignerLeaf[]): Config.Topology
...
+  if (guardTopology) {
+    const leaves = modules.map((module) => {
+      // Leave out the recovery module (which has weight 255)
+      if (module.weight === 255n) {
+        return module
+      }
+
+      return {
+        type: 'nested',
+        weight: 255n,
```
ScreamingHawk, line 461: I don't think this is right. This gives any (module + guard) pair full access to the wallet.

patrislav: With only the sessions module, this works as expected, as (sessions + guard) should be able to access the wallet. But it is gonna break for modules that aren't supposed to have such power. I'll get it changed so that we only nest the sessions and leave other modules as-is

packages/services/guard/src/index.ts
patrislav: The reason we need both is because the way `digest` is derived from `message` depends on what the `message` is, which the guard service cannot know at this point. E.g. when it's TypedData, the message is the JSON but the `digest` is derived from the encoded TypedData. Similarly with attestation (message is the JSON but digest derived from the encoded bytes.)

patrislav: I don't want to move that logic here (and dependency on wallet-primitives) as this should be a dumb service

packages/wallet/wdk/src/sequence/wallets.ts
patrislav: I tried but the amount of work to support every flexible arrangement out there was a bit much. So I did the opposite 😄 We know what the shape of v3 wallet configs should be, we know what the sessions tree looks like (with and without guard), and we already assume a lot about the wallets anyway, so it was only natural to add those assumptions to modules.

If we need more flexibility, we can tackle it then

packages/wallet/wdk/src/sequence/wallets.ts
patrislav: Changed the type of modules from `SapientSignerLeaf` to a new `Module` type that includes guard information. This way we can specify which modules need it and which don't.

packages/services/guard/src/index.ts
ScreamingHawk: Is the guard signing the digest? 
I'd argue it doesn't make sense to include the message for validation if the result it's signing can't be derived. Because a malicious actor could send a valid payload with a malicious digest and guard wouldn't be able to tell.

packages/wallet/wdk/src/sequence/manager.ts
```diff
@@ -106,6 +106,7 @@ export const ManagerOptionsDefaults = {
 
   guardUrl: 'https://dev-guard.sequence.app',
   guardAddress: '0xa2e70CeaB3Eb145F32d110383B75B330fA4e288a' as Address.Address, // TODO: change to the actual guard address
+  guardPrivateKey: '0046e54c861e7d4e1dcd952d86ab6462dedabc55dcf00ac3a99dcce59f516370' as Hex.Hex,
```
ScreamingHawk, line 109: Env var? Developers may end up exposing their dev wallet to malicious actors if they have this known key in the configuration, since the configuration is considered public information.

packages/wallet/dapp-client/src/DappClient.ts
```diff
@@ -246,6 +248,7 @@ export class DappClient {
     this.walletAddress = implicitSession?.walletAddress || explicitSessions[0]?.walletAddress || null
     this.loginMethod = result[0]?.loginMethod || null
     this.userEmail = result[0]?.userEmail || null
+    this.guard = implicitSession?.guard || explicitSessions[0]?.guard
```
ScreamingHawk, line 251: Probably fine but could do `|| explicitSessions.find(s => !!s.guard)?.guard`

packages/wallet/core/src/signers/guard.ts
```diff
@@ -0,0 +1,97 @@
...
+      message: Bytes.fromString(Attestation.toJson(payload.attestation)),
+      digest: Hash.keccak256(Attestation.encode(payload.attestation)),
+    }
+  }
+  const typedData = Payload.toTyped(wallet, chainId, payload)
+  return {
+    message: Bytes.fromString(TypedData.serialize(typedData)),
+    digest: Bytes.fromHex(TypedData.getSignPayload(typedData)),
+  }
+}
```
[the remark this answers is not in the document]
patrislav: The guard service itself does the verification, I don't think we need it here 🤔

PR #860, opened 2025-09-02
Fix sorting of sessions blacklist

Agusx1211:
Checksummed addresses trip the sorting and generate incorrectly sorted arrays.


Agusx1211:
> start:multi:server command is unrelated to this PR
yes

PR #861, opened 2025-09-03
Handle guard 2FA

patrislav:


tolgahan-arikan (approved):

packages/wallet/wdk/src/sequence/handlers/guard.ts
```diff
@@ -30,17 +58,34 @@ export class GuardHandler implements Handler {
...
+                try {
+                  const signature = await this.guard.signEnvelope(request.envelope, { id: e.id, code })
+                  await this.signatures.addSignature(request.id, signature)
+                  resolve(true)
+                } catch (e) {
+                  reject(e)
+                }
+              }
+              await onPromptCode(e.id, respond)
+              resolve(true)
```
Agusx1211, line 83: Should we solve right away? I don't remember this perfectly, but I think `handle` should not return early.

Agusx1211 (approved):
LGTM! Very nice :)

PR #865, opened 2025-09-07
Account: Expose wait for receipt

taylanpince:


Agusx1211 (approved):
LGTM!

VGabriel45:
Will need to add the waitForReceipt argument to the higher level AccountSigner class too before merging @Agusx1211

PR #878, opened 2025-09-17
Implicit sessions are authorized by device signers

ScreamingHawk:
This enables implicit sessions to be signed by device or identity signers. 

This change enables implicit sessions for passkeys. 

The contract still expects exactly 1 identity signer when performing validation. We can get around this by hashing unused identity signers to recover the same image hash but support multiple authorizing signers.


packages/wallet/primitives/src/session-signature.ts
```diff
@@ -117,8 +128,23 @@ export function encodeSessionCallSignatures(
...
+
+  // Check the topology contains exactly one identity signer
+  const identitySigners = getIdentitySigners(topology)
+  if (identitySigner) {
+    if (!identitySigners.some((s) => Address.isEqual(s, identitySigner!))) {
+      throw new Error('Identity signer not found')
+    }
+  } else {
+    // Grab the first one
+    identitySigner = identitySigners[0]!
```
Agusx1211, line 143: I don't think this is a good default, right? it may mask errors working in one-off scenarios

packages/wallet/primitives/src/session-signature.ts
```diff
@@ -117,8 +128,23 @@ export function encodeSessionCallSignatures(
     throw new Error('Incomplete topology')
   }
 
+  if (implicitSigners.length > 0 && !identitySigner) {
+    console.warn('Encodings using implicit signers without identity signer may encode incorrectly')
```
Agusx1211, line 132: Lets just throw here imho

packages/wallet/primitives/src/session-signature.ts
ScreamingHawk: When using explicit signers only it doesn't matter which identity signer we use. We can make this required here but further up the chain we still need to pull a random identity signer for when we aren't using implicits.

ScreamingHawk: Updated

PR #880, opened 2025-09-19
Refactor/types namings tsdoc redundant code

VGabriel45:
This PR introduces the following changes:
- refactor "permissions" to refer to Session Permissions 
- refactor "session" to refer to Session (implicit or explicit)
- refactor "explicitSession" to refer to explicit session
- improved RequestActionType types to cover more explicit scenarios
- added "getAllImplicitSessions()" and getAllExplicitSessions()"
- refactor "permissions" type in wallet/primitives SessionPermissions to be Permission[], [Permission, Permission[]] was causing too many type issues on client side and required to be casted

### DIFF
--- a/packages/wallet/dapp-client/src/types/index.ts
+++ b/packages/wallet/dapp-client/src/types/index.ts
@@ -1,6 +1,6 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import { ExplicitSession } from '@0xsequence/wallet-core'
-import { Attestation, Payload, Permission } from '@0xsequence/wallet-primitives'
+import { Attestation, Payload } from '@0xsequence/wallet-primitives'
 import { Address, Hex } from 'ox'
 import type { TypedData } from 'ox/TypedData'
 
@@ -24,15 +24,16 @@ export interface GuardConfig {
 
 // --- Payloads for Transport ---
 
-export interface AddExplicitSessionPayload {
-  session: ExplicitSession
-  preferredLoginMethod?: LoginMethod
-  email?: string
-}
 export interface CreateNewSessionPayload {
   origin?: string
   session?: ExplicitSession
   includeImplicitSession?: boolean
+  preferredLoginMethod?: LoginMethod
+  email?: string
+}
+
+export interface AddExplicitSessionPayload {
+  session: ExplicitSession
   preferredLoginMethod?: LoginMethod
   email?: string
 }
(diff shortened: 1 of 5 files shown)


packages/wallet/wdk/src/sequence/sessions.ts
```diff
@@ -327,12 +342,12 @@ export class Sessions implements SessionsInterface {
   async addExplicitSession(
     walletAddress: Address.Address,
     sessionAddress: Address.Address,
```
corbanbrook, line 344: sessionAddress is an argument for addExplicitSession but its also a property of the explicitSession.

packages/wallet/primitives/src/permission.ts
```diff
@@ -25,7 +25,7 @@ export type SessionPermissions = {
   chainId: number
   valueLimit: bigint
   deadline: bigint // uint64
-  permissions: [Permission, ...Permission[]]
```
corbanbrook, line 28: This typing [Permission, ...Permission[]] is cute typing trick that uses a tuple type to specify that there must be at least 1 permission given in the array. where as Permission[] allows an empty array. Was this change intended?

packages/wallet/dapp-client/src/types/index.ts
```diff
@@ -141,12 +151,9 @@ export type DappClientWalletActionEventListener = (data: {
   chainId: number
 }) => void
 
-export type DappClientExplicitSessionEventListener = (data: {
-  action: (typeof RequestActionType)['ADD_EXPLICIT_SESSION' | 'MODIFY_EXPLICIT_SESSION']
-  response?: AddExplicitSessionSuccessResponsePayload | ModifySessionSuccessResponsePayload
-  error?: any
+export type DappClientExplicitSessionEventListener = ExplicitSessionEventListener & {
```
corbanbrook, line 154: This is strange to me ... we are tagging the event listener with a chain id?

packages/wallet/dapp-client/src/types/index.ts
```diff
@@ -115,19 +110,34 @@ export type Transaction =
...
+  permissions?: Permission.Permission[]
   chainId?: number
 }
 
+export type ExplicitSession = Omit<Session, 'isImplicit'> & {
+  chainId: number
+  permissions: Permission.Permission[]
+}
+
+export type ImplicitSession = Omit<Session, 'isImplicit'>
```
corbanbrook, line 126: Im not sure why the ImplicitSession type is omitting isImplicit. It might be better to use a discriminating type union here.

VGabriel45: Yep agree, I refactored this one.

PR #888, opened 2025-10-07
add checkForPermissions

VGabriel45:
Refactor "hasPermissions" into "checkForPermissions" to return an extra property "isImplicit" which allows to also check what type of session we are using for the allowed call.


VGabriel45:
Closing this as it's not relevant since we are going to remove some extra checks that were dependent on this change.

VGabriel45:
Closing this as it's not relevant since we are going to remove some extra checks that were dependent on this change.

PR #972, opened 2026-03-02
Cleanup databases after logout

corbanbrook:
New `_cleanupWalletData(wallet)` method (lines 1258–1280):                                                                                                                              
  - Fetches all records from signatures, transactions, messages, and recovery databases in parallel                                                                                     
  - Deletes any records belonging to the logged-out wallet address (using Address.isEqual for case-insensitive comparison)                                                              
  - Best-effort cleanup of auth keys: fetches the wallet's loginTopology, extracts its signers, and deletes their corresponding AuthKey entries                                         
  - Auth key cleanup is wrapped in try/catch so an offline state won't block the logout                                                                                                 
                                                                                                                                                                                        
  Soft logout (skipRemoveDevice: true branch, line 1201): calls _cleanupWalletData after the manager entry and device are removed.

  Hard logout (completeLogout, line 1255): calls _cleanupWalletData after the configuration update is committed and the manager entry/device are removed.

### DIFF
--- a/packages/wallet/wdk/src/sequence/wallets.ts
+++ b/packages/wallet/wdk/src/sequence/wallets.ts
@@ -1198,6 +1198,7 @@ export class Wallets implements WalletsInterface {
         this.shared.databases.manager.del(wallet),
         this.shared.modules.devices.remove(walletEntry.device),
       ])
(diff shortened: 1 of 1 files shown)


corbanbrook:
@codex review

PR #998, opened 2026-05-06
Fix fee options for undeployed wallets

corbanbrook:
Root cause: RpcRelayer.feeOptions was ignoring the caller’s to and always sending to =
  wallet with raw encoded calls, which matches the failing request.

  Changes made:

  - Added Wallet.buildFeeOptionsTransaction() in wallet-core to build the correct fee-
    options target/data.
  - For undeployed wallets, it targets the guest module and prefixes the wallet deploy
    call before wallet.execute.
  - Updated WDK transactions.define() to pass that built target/data into relayer fee
    options.
  - Updated dapp-client fee options to pass the already built signed transaction data.
  - Updated relayer fee-options API to accept optional prebuilt calldata and stop
    overwriting to.
  - Added focused tests for core transaction construction and relayer request body
    behavior.

### DIFF
--- a/packages/services/relayer/src/relayer/rpc-relayer/index.ts
+++ b/packages/services/relayer/src/relayer/rpc-relayer/index.ts
@@ -149,22 +149,24 @@ export class RpcRelayer implements Relayer {
     chainId: number,
     to: Address.Address,
     calls: Payload.Call[],
+    data?: Hex.Hex,
   ): Promise<{ options: FeeOption[]; quote?: FeeQuote }> {
     // IMPORTANT:
     // The relayer FeeOptions endpoint simulates `eth_call(to, data)`.
-    // wallet-webapp-v3 requests FeeOptions with `to = wallet` and `data = Payload.encode(calls, self=wallet)`.
-    // This works for undeployed wallets and avoids guest-module simulation pitfalls.
-    const callsStruct: Payload.Calls = { type: 'call', space: 0n, nonce: 0n, calls: calls }
+    // Callers that already built a wallet transaction should pass its `to` and `data`.
+    // This is required for undeployed wallets because the transaction must target the
+    // guest module and include the deploy call before executing from the wallet.
+    const callsStruct: Payload.Calls = { type: 'call', space: 0n, nonce: 0n, calls }
 
-    const feeOptionsTo = wallet
-    const data = Payload.encode(callsStruct, wallet)
+    const feeOptionsTo = to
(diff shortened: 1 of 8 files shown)


packages/services/relayer/src/relayer/rpc-relayer/index.ts
```diff
@@ -149,22 +149,24 @@ export class RpcRelayer implements Relayer {
...
-    const callsStruct: Payload.Calls = { type: 'call', space: 0n, nonce: 0n, calls: calls }
+    // Callers that already built a wallet transaction should pass its `to` and `data`.
+    // This is required for undeployed wallets because the transaction must target the
+    // guest module and include the deploy call before executing from the wallet.
+    const callsStruct: Payload.Calls = { type: 'call', space: 0n, nonce: 0n, calls }
 
-    const feeOptionsTo = wallet
-    const data = Payload.encode(callsStruct, wallet)
+    const feeOptionsTo = to
+    const feeOptionsData = data ?? Hex.fromBytes(Payload.encode(callsStruct, to))
```
tolgahan-arikan, line 162: The new `data` parameter fixes callers that pass a prebuilt transaction, but the fallback should preserve the #966 behavior for existing four-argument callers. When `data` is absent, this now uses the caller-provided `to` and encodes with `self = to`; if a caller passes the guest/module target without deploy calldata, that can reintroduce the undeployed-wallet simulation issue #966 avoided. I’d keep `data` optional, but branch on it explicitly:

```ts
const hasTransactionData = data !== undefined
const feeOptionsTo = hasTransactionData ? to : wallet
const feeOptionsData = hasTransactionData
  ? data
  : Hex.fromBytes(Payload.encode(callsStruct, wallet))
```

tolgahan-arikan: or we could make `data` non-optional?

tolgahan-arikan: One update after checking `0xsequence/relayer`: preserving the #966 fallback is only useful for older relayer deployments. The newer relayer behavior in [relayer PR #171](https://github.com/0xsequence/relayer/pull/171) checks code at `to`, so `to = wallet` fails for undeployed wallets before simulation. Given that, I think the cleaner fix may be to make `data` required and always pass the prebuilt transaction shape (`to = guest`, `data = deploy + execute` for undeployed wallets).

tolgahan-arikan (approved):

PR #1015, opened 2026-06-20
fix: handle unicode redirect payloads in dapp transport

Ysh204:
## Summary
- Encode redirect request payloads as UTF-8 before base64 encoding
- Decode redirect response payloads as UTF-8 after base64 decoding
- Add regression coverage for Unicode redirect request and response payloads

## Testing
- corepack pnpm --filter @0xsequence/dapp-client exec vitest run
- corepack pnpm --filter @0xsequence/dapp-client lint
- corepack pnpm --filter @0xsequence/dapp-client typecheck
- corepack pnpm --filter @0xsequence/dapp-client build

### DIFF
--- a/.changeset/fresh-redirects-dance.md
+++ b/.changeset/fresh-redirects-dance.md
@@ -1,0 +1,5 @@
+---
+'@0xsequence/dapp-client': patch
+---
+
+Fix redirect transport payload encoding so Unicode characters are handled correctly in redirect requests and responses.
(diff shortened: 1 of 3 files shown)


Ysh204:
Good catch. Updated the decoder to use fatal UTF-8 decoding and fall back to the legacy atob string when the payload is not valid UTF-8. I also added regression coverage for a legacy Latin-1 redirect response payload.

PR #1019, opened 2026-06-22
Expose add login signer redirect metadata

matt416:
Adds an opt-in includeMetadata mode to WDK completeRedirect so add-login-signer redirect callers can receive the new signer address, kind, and OAuth email alongside the redirect target. 

Preserves the default Promise<string> behavior for existing callers and only includes addedLoginSigner for add-signer redirect completions. Adds WDK wallet tests for the metadata response and the legacy string response.

Verified with WDK build, targeted wallets.test.ts, WDK typecheck, and WDK lint

### DIFF
--- a/packages/wallet/wdk/src/sequence/index.ts
+++ b/packages/wallet/wdk/src/sequence/index.ts
@@ -15,6 +15,8 @@ export type {
   MnemonicSignupArgs,
   EmailOtpSignupArgs,
   CompleteRedirectArgs,
+  CompleteRedirectWithMetadataArgs,
+  CompleteRedirectMetadataResult,
   SignupArgs,
   AddLoginSignerArgs,
   RemoveLoginSignerArgs,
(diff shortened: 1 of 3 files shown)


corbanbrook (dismissed):

tolgahan-arikan (dismissed):
Minor comment, not blocker:

The new metadata types are declared in `wallets.ts` but not re-exported from `packages/wallet/wdk/src/sequence/index.ts`. This means consumers can use the feature via inference, but cannot cleanly import/name `CompleteRedirectWithMetadataArgs` or `CompleteRedirectMetadataResult` from the public package entrypoint.

matt416 (commented):
Confirmed — `CompleteRedirectWithMetadataArgs` and `CompleteRedirectMetadataResult` are exported from `wallets.ts` (lines 107 and 111) but not re-exported from `packages/wallet/wdk/src/sequence/index.ts`. The existing type block at line 10–25 of that file re-exports `CompleteRedirectArgs` and the other args types, but the two new metadata types are absent.

Consumers who rely on inference today will be fine, but anyone wanting to name or import either type directly (e.g. in a function signature or a `satisfies` constraint) will need to reach into the internal path or use `ReturnType`/`Parameters` workarounds.

Suggested fix — add the two types to the existing `export type { ... } from './wallets.js'` block:

```ts
CompleteRedirectWithMetadataArgs,
CompleteRedirectMetadataResult,
```

tolgahan-arikan (approved):

